PR #14, opened 2019-02-13
Replace blocking calls by replacing .join() with async implementations.

bainian12345:
- Replace blocking call in HCListener
- Replace blocking call in ClientWebSocket
- Cleanup TimeoutHelper

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -176,7 +176,8 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	CompletableFuture<ByteBuffer> readBinaryAsync(Duration timeout) {
 		LinkedList<byte[]> fragments = new LinkedList<byte[]>();
 		AtomicInteger messageSize = new AtomicInteger(0);
-		return readFragmentsAsync(fragments, messageSize, timeout).thenApply((nullResult) -> {
+		TimeoutHelper timeoutHelper = new TimeoutHelper(timeout, true);
+		return readFragmentsAsync(fragments, messageSize, timeoutHelper).thenApply((nullResult) -> {
 			byte[] message = new byte[messageSize.get()];
 			int offset = 0;
 			for (byte[] bytes : fragments) {
@@ -188,9 +189,8 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 		});
 	}
 
-	private CompletableFuture<Void> readFragmentsAsync(LinkedList<byte[]> fragments, AtomicInteger messageSize, Duration timeout) {
-		TimeoutHelper timeoutHelper = (timeout != null) ? new TimeoutHelper(timeout, true) : null;
-		return fragmentQueue.dequeueAsync(timeout).thenCompose(fragment -> {
+	private CompletableFuture<Void> readFragmentsAsync(LinkedList<byte[]> fragments, AtomicInteger messageSize, TimeoutHelper helper) {
+		return fragmentQueue.dequeueAsync(helper.remainingTime()).thenCompose(fragment -> {
 			if (fragment == null) {
 				return CompletableFuture.completedFuture(null);
 			}
@@ -199,11 +199,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 			fragments.add(fragment.getBytes());
 
 			if (!fragment.isEnd()) {
-				if (timeout != null) {
-					return readFragmentsAsync(fragments, messageSize, timeoutHelper.remainingTime());
-				} else {
-					return readFragmentsAsync(fragments, messageSize, null);
-				}
+				return readFragmentsAsync(fragments, messageSize, helper);
 			}
 			return CompletableFuture.completedFuture(null);
 		});
@@ -306,7 +302,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	 * @return Returns a CompletableFuture which completes when the connection is completely closed.
 	 */
 	CompletableFuture<Void> closeAsync(CloseReason reason) {
-		RelayLogger.logEvent("closing", this);
+		RelayLogger.logEvent("clientWebSocketClosing", this, (reason != null) ? reason.getReasonPhrase() : "NONE");
 		this.fragmentQueue.shutdown();
 		this.textQueue.shutdown();
 		
@@ -352,7 +348,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	
 	@OnClose
 	public void onClose(Session session, CloseReason reason) {
-		RelayLogger.logEvent("closed", this);
+		RelayLogger.logEvent("clientWebSocketClosed", this, reason.getReasonPhrase());
 		this.textQueue.shutdown();
 		this.fragmentQueue.shutdown();
 		
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -6,6 +6,7 @@ import java.time.Duration;
 import java.util.LinkedList;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.io.IOException;
 import java.io.Writer;
 
@@ -173,36 +174,41 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	 * @throws TimeoutException thrown when a complete message frame is not received within the timeout.
 	 */
 	CompletableFuture<ByteBuffer> readBinaryAsync(Duration timeout) {
-		
-		return CompletableFutureUtil.timedSupplyAsync(timeout, () -> {
-				LinkedList<byte[]> fragments = new LinkedList<byte[]>();
-				boolean receivedWholeMsg;
-				int messageSize = 0;
-				
-				do {
-					MessageFragment fragment = fragmentQueue.dequeueAsync().join();
-					if (fragment == null) {
-						break;
-					}				
-					messageSize += fragment.getBytes().length;
-					fragments.add(fragment.getBytes());
-					receivedWholeMsg = fragment.isEnd();
-				} 
-				while (!receivedWholeMsg);
-	
-				byte[] message = new byte[messageSize];
-				int offset = 0;
-				for (byte[] bytes : fragments) {
-					System.arraycopy(bytes, 0, message, offset, bytes.length);
-					offset += bytes.length;
-				}
-	
-				RelayLogger.logEvent("receivedBytes", this, String.valueOf(message.length));
-				return ByteBuffer.wrap(message);
-			},
-			this.executor);
+		LinkedList<byte[]> fragments = new LinkedList<byte[]>();
+		AtomicInteger messageSize = new AtomicInteger(0);
+		return readFragmentsAsync(fragments, messageSize, timeout).thenApply((nullResult) -> {
+			byte[] message = new byte[messageSize.get()];
+			int offset = 0;
+			for (byte[] bytes : fragments) {
+				System.arraycopy(bytes, 0, message, offset, bytes.length);
+				offset += bytes.length;
+			}
+			RelayLogger.logEvent("receivedBytes", this, String.valueOf(message.length));
+			return ByteBuffer.wrap(message);
+		});
 	}
-	
+
+	private CompletableFuture<Void> readFragmentsAsync(LinkedList<byte[]> fragments, AtomicInteger messageSize, Duration timeout) {
+		TimeoutHelper timeoutHelper = (timeout != null) ? new TimeoutHelper(timeout, true) : null;
+		return fragmentQueue.dequeueAsync(timeout).thenCompose(fragment -> {
+			if (fragment == null) {
+				return CompletableFuture.completedFuture(null);
+			}
+
+			messageSize.getAndAdd(fragment.getBytes().length);
+			fragments.add(fragment.getBytes());
+
+			if (!fragment.isEnd()) {
+				if (timeout != null) {
+					return readFragmentsAsync(fragments, messageSize, timeoutHelper.remainingTime());
+				} else {
+					return readFragmentsAsync(fragments, messageSize, null);
+				}
+			}
+			return CompletableFuture.completedFuture(null);
+		});
+	}
+
 	/**
 	 * Sends the data to the remote endpoint as binary.
 	 * 
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -720,7 +720,7 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 		 * @throws InterruptedException
 		 */
 		private CompletableFuture<Void> receivePumpAsync() {
-			return CompletableFuture.supplyAsync(() -> receivePumpCore()).handle((keepGoing, ex) -> {
+			return receivePumpCoreAsync().handle((keepGoing, ex) -> {
 				if (keepGoing) {
 					receivePumpAsync();
 				}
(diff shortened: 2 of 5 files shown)


dlstucki:
> 							new CloseReason(CloseReason.CloseCodes.NORMAL_CLOSURE, "Client closing the socket normally")));

I don't know. It would be nice to ensure there's some way for users to do it.

---
In reply to: [463354662](https://github.com/Azure/azure-relay-java/pull/14#issuecomment-463354662) [](ancestors = 463354662)

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java:312 in 479de6c. [](commit_id = 479de6ca5e0e3c54c1532c09eb383f26ce4184aa, deletion_comment = True)

src/main/java/com/microsoft/azure/relay/TimeoutHelper.java
```diff
@@ -9,11 +9,11 @@
 	private boolean deadlineSet;
 	private Duration originalTimeout;
 
-	protected TimeoutHelper(Duration timeout) {
+	TimeoutHelper(Duration timeout) {
```
dlstucki: >TimeoutHelper [](start = 1, length = 13)

You don't need to change anything now, but I believe the philosophy with public access on types which are internal is to imagine if the type were made public and someone in another package was going to use the class. What should that "public" API look like? In that hypothetical scenario would they need to call the constructor and some of these methods? Probably.

Again no change needed for now, just food for thought next time implementing an internal class.

dlstucki: Maybe these didn't follow the rule I'm describing in the original C# version. If the were internal there that would explain.

---
In reply to: [257346026](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257346026) [](ancestors = 257346026)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -173,36 +174,41 @@ boolean isOpen() {
...
-	
-				byte[] message = new byte[messageSize];
-				int offset = 0;
-				for (byte[] bytes : fragments) {
-					System.arraycopy(bytes, 0, message, offset, bytes.length);
-					offset += bytes.length;
+		LinkedList<byte[]> fragments = new LinkedList<byte[]>();
+		AtomicInteger messageSize = new AtomicInteger(0);
+		return readFragmentsAsync(fragments, messageSize, timeout).thenApply((nullResult) -> {
+			byte[] message = new byte[messageSize.get()];
```
dlstucki: >get [](start = 41, length = 3)

I was going to say if you have readFragmentsAsync return a CompletableFuture<Integer> you might be able to remove the AtomicInteger here. You're the expert on async java now. I suspect it might make the code harder with the quasi-recursive magic that's being used here. However, returning an Integer would mean having to possibly allocate multiple Integers vs one AtomicInteger. Anyhow, just think about how it might look with CF<Int> instead and probably stick with what you have.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -173,36 +174,41 @@ boolean isOpen() {
...
+				System.arraycopy(bytes, 0, message, offset, bytes.length);
+				offset += bytes.length;
+			}
+			RelayLogger.logEvent("receivedBytes", this, String.valueOf(message.length));
+			return ByteBuffer.wrap(message);
+		});
+	}
+
+	private CompletableFuture<Void> readFragmentsAsync(LinkedList<byte[]> fragments, AtomicInteger messageSize, Duration timeout) {
+		TimeoutHelper timeoutHelper = (timeout != null) ? new TimeoutHelper(timeout, true) : null;
```
dlstucki: >timeoutHelper [](start = 16, length = 13)

You could pass a single TimeoutHelper just like you pass the LinkedList and AtomicInteger. Depends if the timeout applies for the overall operation or for each fragment. I personally think it should be for the overall receive so a single TimeoutHelper might work well.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -173,36 +174,41 @@ boolean isOpen() {
...
+
+			messageSize.getAndAdd(fragment.getBytes().length);
+			fragments.add(fragment.getBytes());
+
+			if (!fragment.isEnd()) {
+				if (timeout != null) {
+					return readFragmentsAsync(fragments, messageSize, timeoutHelper.remainingTime());
+				} else {
+					return readFragmentsAsync(fragments, messageSize, null);
 				}
```
dlstucki: It's largely a personal preference but since this method is already using a trick (which is cool) I would try to keep the number of possible re-entry points as little as possible. Look at this alternate then use whichever you prefer.

    Duration remainingTime = null; // Or use ? : 
    if (timeout != null) {
        remainingTime = timeoutHelper.remainingTime();
    }
    return readFragmentsAsync(fragments, messageSize, remainingTime);

dlstucki: This makes me wonder, in Java-land would it be better if TimeoutHelper allowed a null Duration?

---
In reply to: [257368397](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257368397) [](ancestors = 257368397)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -278,49 +278,39 @@ public HybridConnectionListener(String connectionString, String path) throws URI
...
+
+				RelayLogger.logEvent("closing", this);
+				this.closeCalled = true;
+
+				// If the input queue is empty this completes all pending waiters with null and
+				// prevents any new items being added to the input queue.
+				this.connectionInputQueue.shutdown();
+
+				// Close any unaccepted rendezvous. DequeueAsync won't block since we've called
+				// connectionInputQueue.Shutdown().
```
dlstucki: Did you see this comment?

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
bainian12345: I modified TimeoutHelper so that if a null duration was passed in it would be interpreted as max duration

---
In reply to: [257370849](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257370849) [](ancestors = 257370849,257368397)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -277,50 +276,46 @@ public HybridConnectionListener(String connectionString, String path) throws URI
...
+		})
+		.whenComplete((nullResult, ex) -> {
 			this.connectionInputQueue.dispose();
-		}
-		return this.controlConnection.closeAsync(null);
+			if (ex != null) {
+				throw RelayLogger.throwingException(ex, this);
+			}
+		})
+		.thenCompose(nullResult -> this.controlConnection.closeAsync(null));
```
dlstucki: >null [](start = 63, length = 4)

TODO (can be future): Use TimeoutHelper in this method and pass the remaining timeout to controlConnection.closeAsync.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -277,50 +276,46 @@ public HybridConnectionListener(String connectionString, String path) throws URI
...
+				this.isOnline = false;
+				RelayLogger.logEvent("closed", this);
+			});
+		})
+		.whenComplete((nullResult, ex) -> {
 			this.connectionInputQueue.dispose();
-		}
-		return this.controlConnection.closeAsync(null);
+			if (ex != null) {
+				throw RelayLogger.throwingException(ex, this);
```
dlstucki: >throw RelayLogger.throwingException(ex, this); [](start = 4, length = 46)

Does this throw cause the controlConnection to never get closed?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -277,50 +276,46 @@ public HybridConnectionListener(String connectionString, String path) throws URI
...
+				RelayLogger.logEvent("closing", this);
+				this.closeCalled = true;
+
+				// If the input queue is empty this completes all pending waiters with null and
+				// prevents any new items being added to the input queue.
+				this.connectionInputQueue.shutdown();
+
+				// Close any unaccepted rendezvous. DequeueAsync won't block since we've called
+				// connectionInputQueue.Shutdown().
+				CompletableFuture<?>[] closeTasks = new CompletableFuture<?>[this.connectionInputQueue.getPendingCount()];
```
dlstucki: >CompletableFuture<?>[] closeTasks = new CompletableFuture<?>[this.connectionInputQueue.getPendingCount()]; [](start = 4, length = 106)

I'm just curious, does List make this any prettier?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -277,50 +276,46 @@ public HybridConnectionListener(String connectionString, String path) throws URI
 	 * @return A CompletableFuture which completes when all connections are
 	 *         disconnected with the cloud service
 	 */
+	@SuppressWarnings("unchecked")
```
dlstucki: Can this suppress be on the specific line which requires it?

[http://thefinestartist.com/effective-java/24](http://thefinestartist.com/effective-java/24)
"Always use the SuppressWarnings annotation on the smallest scope possible"

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -277,50 +276,46 @@ public HybridConnectionListener(String connectionString, String path) throws URI
...
+								new CloseReason(CloseReason.CloseCodes.NORMAL_CLOSURE, "Client closing the socket normally"));
+					});
+				}
+				return closeTasks;
+			}
+		}, EXECUTOR)
+		.thenCompose(closeTasks -> {
+			return CompletableFuture.allOf((CompletableFuture<Void>[]) closeTasks).thenRun(() -> {
+				this.isOnline = false;
+				RelayLogger.logEvent("closed", this);
```
dlstucki: >logEvent [](start = 16, length = 8)

Logging "closed" should be after controlConnection.closeAsync completes.

src/main/java/com/microsoft/azure/relay/RelayLogger.java
```diff
@@ -144,6 +144,8 @@ private static boolean init() {
 		map.put("writingBytesFailed", "%s: writing bytes failed.");
 		map.put("receivedBytes", "%s: received bytes from remote. Total length: %s");
 		map.put("receivedText", "%s: received text from remote. Total length: %s");
+		map.put("clientWebSocketClosing", "%s: is closing. Close reason: %s");
```
dlstucki: >clientWebSocketClosing [](start = 11, length = 22)

Are these grouped by purpose (or some concept you have)? If so, great. If not, consider alphabetizing. 

My personal rule in my head is "Try to use some kind of grouping. (It doesn't have to be alphabetical if something else seems to be better)."

src/main/java/com/microsoft/azure/relay/TimeoutHelper.java
```diff
@@ -76,38 +76,33 @@ protected static Instant subtract(Instant time, Duration timeout) {
...
+	Duration remainingTime() {
 		if (!this.deadlineSet) {
 			this.setDeadline();
 			return this.originalTimeout;
-		} else if (this.deadline == Instant.MAX) {
-			return RelayConstants.MAX_DURATION;
 		} else {
-			Duration remaining = Duration.between(Instant.now(), this.deadline);
+			Duration remaining = (this.originalTimeout != null) ? 
+					Duration.between(Instant.now(), this.deadline) : RelayConstants.MAX_DURATION;
```
dlstucki: >MAX_DURATION [](start = 69, length = 12)

I'm super glad this doesn't allocate!

dlstucki:
> 		if (!deadlineSet) {

Is this being removed because a review of all the callers shows they only call when deadlineSet == false?

---
Refers to: src/main/java/com/microsoft/azure/relay/TimeoutHelper.java:96 in 479de6c. [](commit_id = 479de6ca5e0e3c54c1532c09eb383f26ce4184aa, deletion_comment = True)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -173,36 +174,37 @@ boolean isOpen() {
...
+		return fragmentQueue.dequeueAsync(helper.remainingTime()).thenCompose(fragment -> {
+			if (fragment == null) {
+				return CompletableFuture.completedFuture(null);
+			}
+
+			messageSize.getAndAdd(fragment.getBytes().length);
+			fragments.add(fragment.getBytes());
+
+			if (!fragment.isEnd()) {
+				return readFragmentsAsync(fragments, messageSize, helper);
```
dlstucki: >helper [](start = 54, length = 6)

the name `helper` is pretty ambiguous unless I look at its declaration.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: I can't figure out a way to make it any smaller scoped.

---
In reply to: [257777052](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257777052) [](ancestors = 257777052)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: In case of an exception, most likely an IO exception, the ControlConnection probably won't be in a state to be closed normally anyways right?

---
In reply to: [257775403](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257775403) [](ancestors = 257775403)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: Will do.

---
In reply to: [257775074](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257775074) [](ancestors = 257775074)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: See what you think of this PR:
[https://github.com/Azure/azure-relay-java/pull/15](https://github.com/Azure/azure-relay-java/pull/15)

---
In reply to: [258610208](https://github.com/Azure/azure-relay-java/pull/14#discussion_r258610208) [](ancestors = 258610208,257775403)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: CompletableFuture.allof() doesn't take List

---
In reply to: [257775841](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257775841) [](ancestors = 257775841)

dlstucki: fair enough

---
In reply to: [258613192](https://github.com/Azure/azure-relay-java/pull/14#discussion_r258613192) [](ancestors = 258613192,257775841)

src/main/java/com/microsoft/azure/relay/RelayLogger.java
bainian12345: Ok, will make it alphabetical. Currently it was mostly grouped by files.

---
In reply to: [257779837](https://github.com/Azure/azure-relay-java/pull/14#discussion_r257779837) [](ancestors = 257779837)

dlstucki: You don't have to, I was just making the comment: Consider grouping by some heuristic. If files is it, then that's fine.

---
In reply to: [258614069](https://github.com/Azure/azure-relay-java/pull/14#discussion_r258614069) [](ancestors = 258614069,257779837)

bainian12345:
> 		if (!deadlineSet) {

Yes, it's a redundant check here.

---
In reply to: [464821383](https://github.com/Azure/azure-relay-java/pull/14#issuecomment-464821383) [](ancestors = 464821383)

---
Refers to: src/main/java/com/microsoft/azure/relay/TimeoutHelper.java:96 in 479de6c. [](commit_id = 479de6ca5e0e3c54c1532c09eb383f26ce4184aa, deletion_comment = True)

PR #16, opened 2019-02-18
Taking a stab at refactoring ClientWebSocket's readBinary

dlstucki:
Taking a stab at refactoring ClientWebSocket's implementation of gathering fragments into a full 'message'.

I was mainly curious to see if I could find some way to make it easier to understand. I'm curious if you think this refactoring is better or worse that the original.

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -719,12 +719,12 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 				if (keepGoing) {				
 					receivePumpAsync();
 				} else {
-					this.onOffline(ex);
 					if (ex != null) {
-						throw RelayLogger.throwingException(ex, this, TraceLevel.WARNING);
-					}
+						RelayLogger.throwingException(ex, this, TraceLevel.WARNING);
+					}					
+					this.onOffline(ex);
 				}
-				
+
 				return null;
 			});
 		}
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -279,29 +279,29 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 	@SuppressWarnings("unchecked")
 	public CompletableFuture<Void> closeAsync(Duration timeout) {
 		return CompletableFutureUtil.timedSupplyAsync(timeout, () -> {
-			synchronized (this.thisLock) {
-				if (this.closeCalled) {
-					return CompletableFuture.completedFuture(null);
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -716,14 +716,15 @@ public void close() {
...
 		private CompletableFuture<Void> receivePumpAsync() {
 			return receivePumpCoreAsync().handle((keepGoing, ex) -> {
-				if (keepGoing) {
+				if (keepGoing) {				
 					receivePumpAsync();
+				} else {
+					this.onOffline(ex);
+					if (ex != null) {
+						throw RelayLogger.throwingException(ex, this, TraceLevel.WARNING);
+					}
```
dlstucki: Revert these order changes

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -716,14 +716,15 @@ public void close() {
 		 */
 		private CompletableFuture<Void> receivePumpAsync() {
 			return receivePumpCoreAsync().handle((keepGoing, ex) -> {
-				if (keepGoing) {
+				if (keepGoing) {				
 					receivePumpAsync();
```
dlstucki: >receivePumpAsync [](start = 5, length = 16)

Is this truly supposed to fall through and call onOffline?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: This much of the original is indeed correct: log the exception, then call onOffline.

---
In reply to: [257831721](https://github.com/Azure/azure-relay-java/pull/16#discussion_r257831721) [](ancestors = 257831721)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: isn't supposed to?  Exit keepGoing because it's going offline

---
In reply to: [257831842](https://github.com/Azure/azure-relay-java/pull/16#discussion_r257831842) [](ancestors = 257831842)

dlstucki: The thing is line 720 kicks off a reconnect, but doesn't wait for it.

---
In reply to: [257831842](https://github.com/Azure/azure-relay-java/pull/16#discussion_r257831842) [](ancestors = 257831842)

bainian12345: You are absolutely right. Having the else block down there should fix it

---
In reply to: [258567226](https://github.com/Azure/azure-relay-java/pull/16#discussion_r258567226) [](ancestors = 258567226,257831842)

PR #17, opened 2019-02-20
Refine HybridConnectionListener.ControlConnection reconnect logic

dlstucki:
Anywhere something in HCL.ControlConnection needs a webSocket it should use ensureConnectTask then get the ClientWebSocket from that `CompletableFuture<ClientWebSocket>`. Under normal circumstances when it's already connected that task will be complete and will simply give the current webSocket. If the ControlConnection is already in the process of reconnecting and a parallel call is made which needs the webSocket too that parallel call will wait on the same connectAsyncTask which will give the new ClientWebSocket when completed.

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -533,10 +533,9 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 		public CompletableFuture<Void> openAsync(Duration timeout) {
 			// Establish a WebSocket connection right now so we can detect any fatal errors		
 			CompletableFuture<ClientWebSocket> connectTask = this.ensureConnectTask(timeout);
-			return connectTask.thenApply((webSocket) -> {
+			return connectTask.thenAccept((webSocket) -> {
 					this.tokenRenewer.setOnTokenRenewed((token) -> this.onTokenRenewed(token));
 					this.receivePumpAsync(); // This starts pumping but doesn't wait
-					return (Void)null;
 				}).whenComplete(($void, err) -> {
 					if (err != null) {
 						// catch (Throwable err) {
@@ -676,11 +675,10 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 
 				ClientWebSocket webSocket = new ClientWebSocket(this.listener.trackingContext, EXECUTOR);
 				return delayTask.thenCompose(($void) -> {
-					return webSocket.connectAsync(websocketUri, timeout, config).thenApply(($void2) -> webSocket)
-						.whenComplete((ws, err) -> {
-							if (err == null) {
-								this.onOnline();
-							}
+					return webSocket.connectAsync(websocketUri, timeout, config)
+						.thenApply(($void2) -> {
+							this.onOnline();
+							return webSocket;
 						});
 				});
 			}
@@ -701,7 +699,7 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 			
 			return connectTask.thenCompose((webSocket) -> webSocket.closeAsync(reason))
 				.exceptionally((exception) -> {
-					// catch (Throwable exception) {
+					// catch and do not rethrow
 					RelayLogger.handledExceptionAsWarning(exception, this.listener);
 					return null;
 				});
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -466,9 +466,9 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 	 * Connects, maintains, and transparently reconnects this listener's control
 	 * connection with the cloud service.
 	 */
-	final class ControlConnection implements AutoCloseable {
+	final static class ControlConnection implements AutoCloseable {
 		// Retries after 0, 1, 2, 5, 10, 30 seconds
-		final Duration[] CONNECTION_DELAY_INTERVALS = { Duration.ZERO, Duration.ofSeconds(1), Duration.ofSeconds(2),
+		private final static Duration[] CONNECTION_DELAY_INTERVALS = { Duration.ZERO, Duration.ofSeconds(1), Duration.ofSeconds(2),
 				Duration.ofSeconds(5), Duration.ofSeconds(10), Duration.ofSeconds(30) };
 		private final HybridConnectionListener listener;
 		private final URI address;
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -538,26 +536,22 @@ public void setOnlineHandler(BiConsumer<Object, Object[]> onOnline) {
...
+			return connectTask.thenApply((webSocket) -> {
 					this.tokenRenewer.setOnTokenRenewed((token) -> this.onTokenRenewed(token));
-					this.receivePumpAsync();
-					success = true;
-				} finally {
-					if (!success) {
-						this.closeOrAbortWebSocketAsync(connectTask, new CloseReason(CloseCodes.UNEXPECTED_CONDITION,
-								"closing web socket connection because something went wrong trying to connect."));			
+					this.receivePumpAsync(); // This starts pumping but doesn't wait
+					return (Void)null;
```
bainian12345: >(Void) [](start = 12, length = 6)

Is this cast needed?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -538,26 +536,22 @@ public void setOnlineHandler(BiConsumer<Object, Object[]> onOnline) {
...
+				}).whenComplete(($void, err) -> {
+					if (err != null) {
+						// catch (Throwable err) {
+						RelayLogger.throwingException(err, this.listener);
+						CloseReason closeReason = new CloseReason(CloseCodes.UNEXPECTED_CONDITION,
+								"closing web socket connection because something went wrong trying to connect.");
```
bainian12345: >web socket [](start = 17, length = 10)

Do you think saying "control web socket" would be more specific?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: It wouldn't compile without either this cast, specifying `.<Void>.thenApply` or putting it into a local variable of type `CompletableFuture<Void>`.

---
In reply to: [258629544](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258629544) [](ancestors = 258629544)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: Sure, I don't think I actually touched the value of this String.

---
In reply to: [258630907](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258630907) [](ancestors = 258630907)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: That's an indication of maybe .thenApply isn't the best fit here. Will .thenAccept work?

---
In reply to: [258631053](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258631053) [](ancestors = 258631053,258629544)

dlstucki: Ooooooh. Awesome info for me to learn! I'll try it out.

---
In reply to: [258632412](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258632412) [](ancestors = 258632412,258631053,258629544)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -676,34 +679,42 @@ public void setOnlineHandler(BiConsumer<Object, Object[]> onOnline) {
...
-					}
-				}).thenRun(() -> this.onOnline());
+				ClientWebSocket webSocket = new ClientWebSocket(this.listener.trackingContext, EXECUTOR);
+				return delayTask.thenCompose(($void) -> {
+					return webSocket.connectAsync(websocketUri, timeout, config).thenApply(($void2) -> webSocket)
+						.whenComplete((ws, err) -> {
+							if (err == null) {
+								this.onOnline();
+							}
+						});
```
bainian12345: I don't think this .whenComplete block is necessary and the "thenApply" should be enough. If there is an exception from websocket connecting, the "thenApply" wouldn't be run anyways.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -676,34 +679,42 @@ public void setOnlineHandler(BiConsumer<Object, Object[]> onOnline) {
...
-						return this.webSocket.connectAsync(websocketUri, timeout, config);
-					}
-				}).thenRun(() -> this.onOnline());
+				ClientWebSocket webSocket = new ClientWebSocket(this.listener.trackingContext, EXECUTOR);
+				return delayTask.thenCompose(($void) -> {
+					return webSocket.connectAsync(websocketUri, timeout, config).thenApply(($void2) -> webSocket)
+						.whenComplete((ws, err) -> {
+							if (err == null) {
+								this.onOnline();
+							}
```
dlstucki: >} [](start = 7, length = 1)

should I put an explicit throw (with trace) in here?
```
.whenComplete((ws, err) -> {
       if (err == null) {
           this.onOnline();
!>   } else {
!>       throw RelayLogger.throwingException(err);
!>   }
```

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: That sounds good.

---
In reply to: [258635262](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258635262) [](ancestors = 258635262)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: This changed to a thenApply so this comment is out of date.

---
In reply to: [258635516](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258635516) [](ancestors = 258635516)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: Fixed in iteration 3 (iteration 2 was a merge with master).

---
In reply to: [258633157](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258633157) [](ancestors = 258633157,258632412,258631053,258629544)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: Fixed in iteration 3 (iteration 2 was a merge with master).

---
In reply to: [258635798](https://github.com/Azure/azure-relay-java/pull/17#discussion_r258635798) [](ancestors = 258635798,258635262)

bainian12345:
> 		CompletableFuture<?>[] closeTasks;

Just curious why was this declaration explicitly separated from the assignment below?

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java:281 in c9248cd. [](commit_id = c9248cdcca25268d40ea90740a8ffec295638d04, deletion_comment = False)

dlstucki:
> 		CompletableFuture<?>[] closeTasks;

Because the collection is grabbed inside the sync block, but used later outside the sync block.

---
In reply to: [465719611](https://github.com/Azure/azure-relay-java/pull/17#issuecomment-465719611) [](ancestors = 465719611)

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java:281 in c9248cd. [](commit_id = c9248cdcca25268d40ea90740a8ffec295638d04, deletion_comment = False)

bainian12345:
> 		CompletableFuture<?>[] closeTasks;

Oh, I see

---
In reply to: [465720122](https://github.com/Azure/azure-relay-java/pull/17#issuecomment-465720122) [](ancestors = 465720122,465719611)

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java:281 in f693cb4. [](commit_id = f693cb46300fd93f7eabedf0a85af9a5692039b0, deletion_comment = False)

PR #20, opened 2019-02-21
For your consideration. Add a helper method to run code inside AsyncLock/Semaphore

dlstucki:
Add a method which acquires an asyncLock/Semaphore invokes user async code, then releases the lockRelease.

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
+++ b/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
@@ -3,6 +3,8 @@ package com.microsoft.azure.relay;
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Supplier;
 
 public class AsyncSemaphore {
 	private final Object thisLock = new Object();
@@ -89,6 +91,10 @@ public class AsyncSemaphore {
 		});
 	}
 	
+	public <T> CompletableFuture<T> runInsideLockAsync(Duration timeout, AutoShutdownScheduledExecutor executor, Supplier<? extends CompletionStage<T>> supplier) {
+		return new LockScope().runInsideLockAsync(timeout, executor, supplier);
+	}
+	
 	private void release(int count) {
 		synchronized (this.thisLock) {
 			this.permits += count;
@@ -124,6 +130,25 @@ public class AsyncSemaphore {
 				AsyncSemaphore.this.release(count);
 				this.remaining -= count;
 			}
+		}
+	}
+		
+	private final class LockScope {
+		private LockRelease lockRelease;
+		
+		public <T> CompletableFuture<T> runInsideLockAsync(Duration timeout, AutoShutdownScheduledExecutor executor, Supplier<? extends CompletionStage<T>> supplier) {
+			return acquireAsync(timeout, executor)
+				.thenCompose((lockRelease) -> {
+					this.lockRelease = lockRelease;
+					
+					// Invoke the caller supplier function
+					return supplier.get();
+				})
+				.whenComplete(($void, ex) -> {
+					if (this.lockRelease != null) {						
+						this.lockRelease.release();
+					}
+				});
 		}
 	}
 }
--- a/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
+++ b/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
@@ -4,6 +4,7 @@ import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 public class AsyncSemaphore {
@@ -91,8 +92,49 @@ public class AsyncSemaphore {
 		});
 	}
 	
-	public <T> CompletableFuture<T> runInsideLockAsync(Duration timeout, AutoShutdownScheduledExecutor executor, Supplier<? extends CompletionStage<T>> supplier) {
-		return new LockScope().runInsideLockAsync(timeout, executor, supplier);
+	/**
+	  * Acquire the lock without blocking, execute the given asynchronous code, 
+	  * and finally release the lock regardless if the given supplier threw during its execution
+	  * @param timeout Maximum time to wait for the lock. Throws TimeoutException if the lock is not acquired after the timeout finishes
+	  * @param executor The executor that executes the given supplier code
+	  * @param supplier Code to be executed that returns a CompletionStage after obtaining the lock
+	  * @return A CompletableFuture that completes when the lock is released after returning the desired CompletableFuture from the supplier
+	  */
+	public <T> CompletableFuture<T> lockThenCompose(Duration timeout, AutoShutdownScheduledExecutor executor, Supplier<? extends CompletionStage<T>> supplier) {
+		AtomicReference<LockRelease> lockReleaseRef = new AtomicReference<LockRelease>();
+		return acquireAsync(timeout, executor)
+			.thenCompose((lockRelease) -> {
+				lockReleaseRef.set(lockRelease);
+				
+				// Invoke the caller supplier function
+				return supplier.get();
+			})
+			.whenComplete((result, ex) -> {
+				LockRelease lockRelease = lockReleaseRef.get();
+				if (lockRelease != null) {						
+					lockRelease.release();
+				}
+			});
+	}
+	
+	/**
+	  * Acquire the lock without blocking, execute the given synchronous code, 
+	  * and finally release the lock regardless if the given supplier threw during its execution
+	  * @param timeout Maximum time to wait for the lock. Throws TimeoutException if the lock is not acquired after the timeout finishes
+	  * @param executor The executor that executes the given supplier code
+	  * @param supplier Code to be executed that provides a return value after obtaining the lock
+	  * @return A CompletableFuture that completes when the lock is released after returning the desired return value from the supplier
+	  */
+	public <T> CompletableFuture<T> lockThenApply(Duration timeout, AutoShutdownScheduledExecutor executor, Supplier<T> supplier) {
+		return acquireAsync(timeout, executor)
+			.thenApply((lockRelease) -> {
+				try {
+					// Invoke the caller supplier function
+					return supplier.get();
+				} finally {
+					lockRelease.release();
+				}
+			});
 	}
 	
 	private void release(int count) {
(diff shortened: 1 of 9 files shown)


src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
```diff
@@ -89,6 +91,10 @@ int availablePermits() {
 		});
 	}
 	
+	public <T> CompletableFuture<T> runInsideLockAsync(Duration timeout, AutoShutdownScheduledExecutor executor, Supplier<? extends CompletionStage<T>> supplier) {
```
dlstucki: >runInsideLockAsync [](start = 33, length = 18)

I'm not attached to this name. Please make any suggestions you have for better names.

dlstucki: A few I could think of:

- lockThenCompose[Async]
- lockThenRun[Async]
- runInsideLock[Async]

---
In reply to: [259040952](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259040952) [](ancestors = 259040952)

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
```diff
@@ -119,4 +121,91 @@ public void lockTimeoutTest() throws InterruptedException, ExecutionException, T
...
+		
+		CompletableFuture<String> innerTask = new CompletableFuture<String>();
+		CompletableFuture<String> outerTask = lock.runInsideLockAsync(null, EXECUTOR, () -> {
+			assertTrue("Should be locked during Inner Function", lock.isLocked());
+			return innerTask;
+		});
+		
+		assertFalse(outerTask.isDone());
+		assertTrue("Should be locked now", lock.isLocked());
+		innerTask.complete(originalString);
```
bainian12345: >innerTask.complete(originalString); [](start = 2, length = 35)

Should we check if outerTask is complete here before we manually call it to complete?

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
```diff
@@ -119,4 +121,91 @@ public void lockTimeoutTest() throws InterruptedException, ExecutionException, T
...
+		CompletableFuture<String> innerTaskResult = new CompletableFuture<String>();
+		CompletableFuture<String> outerTask = lock.runInsideLockAsync(null, EXECUTOR, () -> {
+			assertTrue("Should be locked during Inner Function", lock.isLocked());
+			innerTaskStart.complete(null);			
+			return innerTaskResult;
+		});
+		
+		assertFalse(outerTask.isDone());
+		assertTrue("Should be locked now", lock.isLocked());
+		innerTaskResult.complete(originalString);
```
bainian12345: >innerTaskResult.complete(originalString); [](start = 2, length = 41)

So what's the purpose of completing it here? We can just return a completed future from inside directly. I'm not sure if instead of manually completing it every time, there will be more value in replacing it with a little function that sleeps or something to resemble more real life scenarios.

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
dlstucki: line 137 asserts that outerTask is not done. Is that what you're looking for? Or do you want the test to wait a little bit like this:

  Throwable error = assertThrows(ExecutionException.class, (Executable)() -> outerTask.get(TIMEOUT_MS, TimeUnit.MILLISECONDS));
  assertEquals(TimeoutException.class, error.getCause().getClass());

---
In reply to: [259048928](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259048928) [](ancestors = 259048928)

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
bainian12345: I feel like we should have "lock", "compose", and  "release"  all somehow jammed together, but that would be a long and awkward name. I'm ok with "lockThenComposeAsync"

---
In reply to: [259043911](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259043911) [](ancestors = 259043911,259040952)

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
bainian12345: I meant like outerTask should be automatically completed right after we call innerTask.complete(). But that's beside the point of the test I guess

---
In reply to: [259059280](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259059280) [](ancestors = 259059280,259048928)

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
dlstucki: The term 'inside' was meant to indicate it acquires, runs whatever code inside, and then releases.

---
In reply to: [259060038](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259060038) [](ancestors = 259060038,259043911,259040952)

dlstucki: If you like lockThenComposeAsync I'll use that.

---
In reply to: [259061669](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259061669) [](ancestors = 259061669,259060038,259043911,259040952)

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
dlstucki: Isn't what you're describing exactly what the test is doing? After calling innerTask.complete the next line gives outerTask TIMEOUT_MS to complete otherwise get throws and the test fails.

---
In reply to: [259060993](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259060993) [](ancestors = 259060993,259059280,259048928)

src/test/java/com/microsoft/azure/relay/TestUtil.java
```diff
@@ -9,4 +10,23 @@
 	public static final String ENTITY_PATH = CONNECTION_STRING_BUILDER.getEntityPath();
 	public static final String KEY_NAME = CONNECTION_STRING_BUILDER.getSharedAccessKeyName();
 	public static final String KEY = CONNECTION_STRING_BUILDER.getSharedAccessKey();
+		
+	@SuppressWarnings("unchecked")
+	public static <T extends Throwable> T assertThrows(Class<T> expectedType, Executable executable) {
```
bainian12345: >assertThrows [](start = 39, length = 12)

Do you mind adding a bit of documentation to the function?

 /**
  * Asserts that the execution of the given Executable will throw the expected exception
  * @param expectedType The class of the expected exception
  * @param executable The code to be executed
  * @return The actual exception that's thrown, if it matches the expected exception type
  */

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
bainian12345: What do you think of "synchronizedAsync", since the default synchronized keyword implies lock and release inside the block?

---
In reply to: [259061786](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259061786) [](ancestors = 259061786,259061669,259060038,259043911,259040952)

bainian12345: >AutoShutdownScheduledExecutor [](start = 70, length = 29)

Maybe use ExecutorService instead here?

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
```diff
@@ -89,6 +91,10 @@ int availablePermits() {
 		});
 	}
```
bainian12345: Let's add some documentation for this function

 /**
  * Acquire the lock without blocking, then execute the given code, 
  * and finally releases the lock regardless if the given supplier threw during its execution
  * @param timeout Maximum time to wait for the lock. Throws TimeoutException if the lock is not acquired after the timeout finishes
  * @param executor The executor that executes the given supplier code
  * @param supplier Code to be executed that returns a CompletionStage after obtining the lock
  * @return A CompletableFuture that completes when the lock is released after returning the desired CompletableFuture from the supplier
  */

dlstucki: Good suggestion and I really appreciate your providing it :)

---
In reply to: [259079315](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259079315) [](ancestors = 259079315)

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
dlstucki: I wanted the test to create outerTask but not have innerTaskResult completed yet. But since this test blocks on the AsyncLock already being locked I could simplify this. Is that what you're asking for?

---
In reply to: [259058136](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259058136) [](ancestors = 259058136)

dlstucki: Simplifying this in the next iteration.

---
In reply to: [259453912](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259453912) [](ancestors = 259453912,259058136)

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
dlstucki: I'll look into it. The Problem is what type does AsyncSemaphore.acquireAsync take?

---
In reply to: [259064026](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259064026) [](ancestors = 259064026)

dlstucki: I'd have to change the parameters for AsyncSemaphore and InputQueue. Whee!

---
In reply to: [259476726](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259476726) [](ancestors = 259476726,259064026)

bainian12345: That's fine then, not a big deal.

---
In reply to: [259477433](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259477433) [](ancestors = 259477433,259476726,259064026)

dlstucki: lockThenCompose and lockThenApply have been settled upon.

---
In reply to: [259063410](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259063410) [](ancestors = 259063410,259061786,259061669,259060038,259043911,259040952)

src/test/java/com/microsoft/azure/relay/TestUtil.java
dlstucki: Function moved to Assertions.java and documentation added there.

---
In reply to: [259062359](https://github.com/Azure/azure-relay-java/pull/20#discussion_r259062359) [](ancestors = 259062359)

PR #22, opened 2019-02-22
Change AsyncSemaphore to use public interface ScheduledExecutorService

dlstucki:
This required changing InputQueue and a few other things to use the interface.

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
+++ b/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
@@ -8,14 +8,14 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
-public class AsyncSemaphore {
+class AsyncSemaphore {
 	private final Object thisLock = new Object();
 	private final ScheduledExecutorService executor;
 	private final int limit;
 	private InputQueue<Boolean> waiterQueue;
 	private int permits;
 	
-	AsyncSemaphore(int permits, ScheduledExecutorService executor) {
+	public AsyncSemaphore(int permits, ScheduledExecutorService executor) {
 		this.limit = permits;
 		this.executor = executor;
 		synchronized(this.thisLock) {
--- a/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
+++ b/src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
@@ -4,17 +4,20 @@ import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 public class AsyncSemaphore {
 	private final Object thisLock = new Object();
+	private final ScheduledExecutorService executor;
 	private final int limit;
 	private InputQueue<Boolean> waiterQueue;
 	private int permits;
 	
-	AsyncSemaphore(int permits) {
+	AsyncSemaphore(int permits, ScheduledExecutorService executor) {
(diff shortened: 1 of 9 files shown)


src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
```diff
@@ -206,4 +206,28 @@ public void lockScopeUncontendedThrow() throws Throwable {
...
+	@Test
+	public void lockScopeContendedTimeout() throws Throwable {
+		AsyncLock lock = new AsyncLock(EXECUTOR);	
+		LockRelease lockRelease = lock.acquireAsync().join();
+		
+		CompletableFuture<Void> innerTaskStart = new CompletableFuture<Void>();
+		CompletableFuture<String> outerTask = lock.lockThenCompose(Duration.ofMillis(TIMEOUT_MS), () -> {
+			assertTrue("Should be locked during Inner Function", lock.isLocked());
+			innerTaskStart.complete(null);			
+			throw new CompletionException(new IOException("Test Induced Exception"));
```
dlstucki: >throw new CompletionException(new IOException("Test Induced Exception")); [](start = 3, length = 73)

This line never runs, not truly needed.

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
```diff
@@ -4,17 +4,20 @@
...
 
 public class AsyncSemaphore {
 	private final Object thisLock = new Object();
+	private final ScheduledExecutorService executor;
 	private final int limit;
 	private InputQueue<Boolean> waiterQueue;
 	private int permits;
 	
-	AsyncSemaphore(int permits) {
+	AsyncSemaphore(int permits, ScheduledExecutorService executor) {
```
dlstucki: >executor [](start = 54, length = 8)

It's better to put this in the .ctor because when the other calls were providing their own they might have actually used one from a previous call. Better to be consistent instead of misleading.

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
```diff
@@ -33,19 +36,19 @@ int availablePermits() {
 		}
 	}
 
-	CompletableFuture<LockRelease> acquireAsync(AutoShutdownScheduledExecutor executor) {
```
bainian12345: >CompletableFuture<LockRelease> acquireAsync [](start = 1, length = 43)

maybe we should make these acquire methods public

src/test/java/com/microsoft/azure/relay/AsyncLockTest.java
bainian12345: Does anything in here run? If not, we can probably omit all of it except maybe assert the inside isn't reached.

---
In reply to: [259493522](https://github.com/Azure/azure-relay-java/pull/22#discussion_r259493522) [](ancestors = 259493522)

src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java
dlstucki: They definitely should be. I'm a bit unclear if the decided names are supposed to be acquireAsync or lockAsync since both occur.

---
In reply to: [259560190](https://github.com/Azure/azure-relay-java/pull/22#discussion_r259560190) [](ancestors = 259560190)

bainian12345: I would prefer acquireAsync to be similar to the Java API

---
In reply to: [259912209](https://github.com/Azure/azure-relay-java/pull/22#discussion_r259912209) [](ancestors = 259912209,259560190)

dlstucki:
> 			release(1);

it wasn't super clear whether this was calling LockRelease.release(int) or AsyncSemaphore.this.release(int).

---
Refers to: src/main/java/com/microsoft/azure/relay/AsyncSemaphore.java:164 in 83161ee. [](commit_id = 83161ee4f5ab1dfc4e0887dc9bdec7292ccba7cf, deletion_comment = True)

PR #23, opened 2019-02-25
Preview preparations

bainian12345:
- Use slf4j logger instead of log4j
- Modify some API docs to build with Maven
- Modify pom.xml so the project is no longer built with all the dependencies


src/main/java/com/microsoft/azure/relay/RelayLogger.java
```diff
@@ -107,7 +105,7 @@ private static String createSourceString(Object source) {
 	}
 
 	private static void init() {
-		DOMConfigurator.configure("resources/log4j.xml");
+//		DOMConfigurator.configure("resources/log4j.xml");
```
dlstucki: Should this be removed or is there some reason to keep it in here?

dlstucki: >log4j.xml [](start = 41, length = 9)

Is the log4j.xml file now garbage? If so, should it be deleted?

dlstucki: Or should this go into the test package?

---
In reply to: [260062446](https://github.com/Azure/azure-relay-java/pull/23#discussion_r260062446) [](ancestors = 260062446)

PR #24, opened 2019-02-26
Terminate running websocket instances after closing

bainian12345:
Terminate running websocket instances after closing the session

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -12,6 +12,7 @@ import java.io.Writer;
 import javax.websocket.*;
 
 import org.eclipse.jetty.io.RuntimeIOException;
+import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.websocket.api.UpgradeException;
 
 class ClientWebSocket extends Endpoint implements RelayTraceSource {
@@ -288,6 +289,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 			} else {
 				this.session.close();
 			}
+			((LifeCycle) this.container).stop();
 		} 
 		catch (Throwable e) {
 			this.closeTask.completeExceptionally(e);
@@ -326,11 +328,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 		this.fragmentQueue.shutdown();
 		
 		this.closeReason = reason;
-		if (reason.getCloseCode() != CloseReason.CloseCodes.NORMAL_CLOSURE) {
-			this.closeTask.completeExceptionally(new RuntimeIOException("Websocket did not close properly: " + reason.toString()));
-		} else {
-			this.closeTask.complete(null);
-		}
+		this.closeTask.complete(null);
 	}
 
 	@OnError
--- a/src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
+++ b/src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
@@ -165,9 +165,7 @@ public class HybridConnectionListenerTest {
 		CompletableFuture<?>[] clientCloseFutures = new CompletableFuture<?>[MAX_CONNECTIONS_COUNT];
 		for (int i = 0; i < MAX_CONNECTIONS_COUNT; i++) {
 			HybridConnectionChannel clientConnection = (HybridConnectionChannel) clientConnectFutures[i].join();
-			clientCloseFutures[i] = clientConnection.closeAsync()
-				.whenComplete((result, ex) -> {
-				});
+			clientCloseFutures[i] = clientConnection.closeAsync();
 		}
 		
 		CompletableFuture.allOf(clientCloseFutures).join();
--- a/src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
+++ b/src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
@@ -38,7 +38,7 @@ public class HybridConnectionListenerTest {
 	@AfterClass
 	public static void cleanup() {
 		if (listener != null) {
-			listener.closeAsync().join();
+			listener.close();
 		}
 	}
 	
@@ -98,7 +98,6 @@ public class HybridConnectionListenerTest {
 		// Set up listener to accept MAX_CONNECTIONS_COUNT connections async
 		CompletableFuture<?>[] listenerConnectFutures = new CompletableFuture<?>[MAX_CONNECTIONS_COUNT];
 		for (int i = 0; i < MAX_CONNECTIONS_COUNT; i++) {
-			int idx = i;
 			listenerConnectFutures[i] = listener.acceptConnectionAsync()
 				.thenApply(listenerConnection -> {
 					//System.out.println(Thread.currentThread().getName() + " Listener connection " + idx + " " + listenerConnection.getTrackingContext().getTrackingId() + " accepted");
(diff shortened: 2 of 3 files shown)


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -326,11 +328,7 @@ public void onClose(Session session, CloseReason reason) {
 		this.fragmentQueue.shutdown();
 		
 		this.closeReason = reason;
-		if (reason.getCloseCode() != CloseReason.CloseCodes.NORMAL_CLOSURE) {
-			this.closeTask.completeExceptionally(new RuntimeIOException("Websocket did not close properly: " + reason.toString()));
-		} else {
-			this.closeTask.complete(null);
-		}
```
bainian12345: Having the "((LifeCycle) this.container).stop();" line above will sometimes cause the close reason to be ones that's not NORMAL_CLOSURE (such as DISCONNECT), so taking out the throw to avoid problems

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -164,8 +161,6 @@ public void connectMultipleClientsTest() throws URISyntaxException {
 		CompletableFuture<?>[] listenerCloseFutures = new CompletableFuture<?>[MAX_CONNECTIONS_COUNT];
 		for (int i = 0; i < MAX_CONNECTIONS_COUNT; i++) {
 			HybridConnectionChannel listenerConnection = (HybridConnectionChannel) listenerConnectFutures[i].join();
-			int idx = i;
```
dlstucki: >idx [](start = 7, length = 3)

These were used in conjunction with the commented out tracing. Maybe we should get rid of that too.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -38,7 +38,7 @@ public static void init() throws URISyntaxException, RelayException {
 	@AfterClass
 	public static void cleanup() {
```
dlstucki: >cleanup [](start = 20, length = 7)

It _really_ bothers me that there's not a single test case which closes a listener. We could be missing a bunch of shutdown issues.

dlstucki: I'd be quite tempted to have each test case start/stop their own listener and have many/most of those tests fail if stopping throws any exception (e.g. no catch block around listener.close).

---
In reply to: [260847718](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260847718) [](ancestors = 260847718)

dlstucki: Or maybe there could be another class with some but not all of these test cases which does open/close on a per-test basis?

---
In reply to: [260848343](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260848343) [](ancestors = 260848343,260847718)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
dlstucki: Does this result in ever hiding relevant information from the user?

---
In reply to: [260548769](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260548769) [](ancestors = 260548769)

bainian12345: There is the logging line above that records the close reason, so even if something unexpected happened it should be recorded

---
In reply to: [260851931](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260851931) [](ancestors = 260851931,260548769)

dlstucki: As a general point (I'm not sure that it applies here and am not suggesting there's further work for this PR) if we have info that users may want to use to make a decision at runtime then we should give it to them.

In this case if there's nothing they might want to do different depending on whether it was graceful close or not then it may be fine to not pass this info along.

---
In reply to: [260866112](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260866112) [](ancestors = 260866112,260851931,260548769)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: I agree, I will try to come up with some tests for closeAsync() in another PR

---
In reply to: [260849007](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260849007) [](ancestors = 260849007,260848343,260847718)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
dlstucki: Also, customers rarely gather traces from what I've observed. :(

---
In reply to: [260867220](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260867220) [](ancestors = 260867220,260866112,260851931,260548769)

dlstucki: Ugh, I just realized which class this is. Customers of ClientWebSocket may indeed want to know what happened at the protocol level when they requested close happen a certain way and the remote server didn't cooperate. I could see an argument that the users of ClientWebSocket might want the details if close wasn't graceful. 

However, users of the HybridConnection API are less likely to need details if they asked for a close and something happened less than gracefully. Since they don't need to act differently, like call some abort method, then they don't necessarily need to know about it.

---
In reply to: [260868130](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260868130) [](ancestors = 260868130,260867220,260866112,260851931,260548769)

bainian12345: What if we pass the CloseReason as the result to the closeTask? so closeAsync() returns CF<CloseReason> instead of CF<Void>?

---
In reply to: [260867220](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260867220) [](ancestors = 260867220,260866112,260851931,260548769)

dlstucki: I think that would be great!

---
In reply to: [260870149](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260870149) [](ancestors = 260870149,260867220,260866112,260851931,260548769)

dlstucki: Actually, it's already available through getCloseReason!

---
In reply to: [260871743](https://github.com/Azure/azure-relay-java/pull/24#discussion_r260871743) [](ancestors = 260871743,260870149,260867220,260866112,260851931,260548769)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -164,28 +155,18 @@ public void connectMultipleClientsTest() throws URISyntaxException {
...
 
 		// Close client side connections async
 		CompletableFuture<?>[] clientCloseFutures = new CompletableFuture<?>[MAX_CONNECTIONS_COUNT];
 		for (int i = 0; i < MAX_CONNECTIONS_COUNT; i++) {
 			HybridConnectionChannel clientConnection = (HybridConnectionChannel) clientConnectFutures[i].join();
-			int idx = i;
-			String trackingId = clientConnection.getTrackingContext().getTrackingId();
-			//System.out.println(idx + " Sender connection " + trackingId + " closing");						
 			clientCloseFutures[i] = clientConnection.closeAsync()
 				.whenComplete((result, ex) -> {
```
dlstucki: >whenComplete [](start = 5, length = 12)

I suppose this one could go away too

PR #25, opened 2019-02-27
change implementation of listener.isOnline()

bainian12345:


src/main/java/com/microsoft/azure/relay/CompletableFutureUtil.java
```diff
@@ -81,4 +81,8 @@ public T call() throws Exception {
 		}
 		return taskCF;
 	}
+	
+	static boolean isDoneNormally(CompletableFuture<?> cf) {
```
dlstucki: >isDoneNormally [](start = 16, length = 14)

Since this is for callers outside this class I'd put it higher in the file than any private methods.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -758,12 +748,11 @@ public void close() {
 
 		private void onOnline() {
 			synchronized (this.thisLock) {
-				if (this.isOnline) {
+				if (this.isOnline()) {
```
dlstucki: >isOnline [](start = 13, length = 8)

Is it ok to reacquire the same synchonized lock in Java? (This method synchronizes on thisLock and calls isOnline which also then synchronizes on thisLock).

dlstucki:
> 	private final static class ControlConnection implements AutoCloseable {

Since you were talking about test code having access to this class, can you think of a way to write a test for this ControlConnection class which does something like:

- Create and Open ControlConnection
- Verify isOnline
- Find someway to break the websocket connection (may require some kind of test hook)
- Verify onConnecting got called and at that time isOnline == false
- Once reconnected ensure isOnline == true
- Close
- Verify isOffline == true.

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java:450 in 186e0aa. [](commit_id = 186e0aa7dc883b4fa779833fa902a284f3a0e663, deletion_comment = False)

bainian12345:
> 	private final static class ControlConnection implements AutoCloseable {

Yes, I'm working on that. This PR is a prelude towards those tests

---
In reply to: [468027940](https://github.com/Azure/azure-relay-java/pull/25#issuecomment-468027940) [](ancestors = 468027940)

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java:450 in 186e0aa. [](commit_id = 186e0aa7dc883b4fa779833fa902a284f3a0e663, deletion_comment = False)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: If you are worried about it creating a deadlock, I have tested it nd it won't happen.

---
In reply to: [260939178](https://github.com/Azure/azure-relay-java/pull/25#discussion_r260939178) [](ancestors = 260939178)

src/main/java/com/microsoft/azure/relay/CompletableFutureUtil.java
```diff
@@ -8,14 +8,15 @@
...
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 final class CompletableFutureUtil {
-	static AtomicInteger tasksRunning = new AtomicInteger(0);
-
-	static CompletableFuture<Void> delayAsync(Duration delay, ScheduledExecutorService executor) {
+	public static boolean isDoneNormally(CompletableFuture<?> cf) {
```
dlstucki: >isDoneNormally [](start = 23, length = 14)

Personally I might name this "isCompletedNormally" to be similar to "isCompletedExceptionally".

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: good that it won't deadlock (the lock is re-entrant). I general I try to avoid asking for a lock while already owning that lock.

---
In reply to: [260945799](https://github.com/Azure/azure-relay-java/pull/25#discussion_r260945799) [](ancestors = 260945799,260939178)

PR #26, opened 2019-02-27
Cleanup listener handlers

bainian12345:


src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -111,6 +113,57 @@ public void setMaxWebSocketBufferSize(int maxWebSocketBufferSize) {
 		}
 	}
 
+	/**
+	 * Returns the handler that will be run when the listener disconnects unexpectedly due to potential error.
```
dlstucki: >due to potential error [](start = 84, length = 22)

The "due to potential error" doesn't seem needed to me. If the listener disconnects unexpectedly it's an error in my mind.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -755,10 +780,10 @@ private void onOnline() {
 				this.lastError = null;
 				this.connectDelayIndex = -1;
 			}
-
 			RelayLogger.logEvent("connected", this.listener);
-			if (this.onlineHandler != null) {
-				this.onlineHandler.accept(this, null);
+			
+			if (this.listener.getOnlineHandler() != null) {
+				this.listener.getOnlineHandler().run();
```
dlstucki: this pattern is not threadsafe. In cases like this the null check must check something that cannot become null between the check and the usage.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -767,6 +792,10 @@ private void onOffline(Throwable lastError) {
 				this.lastError = lastError;
 			}
 			RelayLogger.logEvent("offline", this);
+			
+			if (this.listener.getOfflineHandler() != null) {
```
dlstucki: >getOfflineHandler [](start = 21, length = 17)

Make this threadsafe.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -783,9 +812,10 @@ private boolean onDisconnect(Exception lastError) {
 			// Inspect the close status/description to see if this is a terminal case
 			// or we should attempt to reconnect.
 			boolean shouldReconnect = this.shouldReconnect(lastError);
+			RelayLogger.logEvent("disconnect", this, String.valueOf(shouldReconnect));
 
-			if (shouldReconnect && this.connectingHandler != null) {
-				this.connectingHandler.accept(this, null);
+			if (shouldReconnect && this.listener.getConnectingHandler() != null) {
```
dlstucki: >this.listener.getConnectingHandler() [](start = 26, length = 36)

Make this threadsafe.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -111,6 +113,57 @@ public void setMaxWebSocketBufferSize(int maxWebSocketBufferSize) {
 		}
 	}
 
+	/**
+	 * Returns the handler that will be run when the listener disconnects unexpectedly due to potential error.
+	 * The listener will attempt to reconnect after this handler runs.	
+	 */
+	public Consumer<Throwable> getConnectingHandler() {
+		return connectingHandler;
+	}
```
dlstucki: Since this is touching a bunch of properties would you consider putting the set/get "properties" after the constructors?

bainian12345: I should probably make another PR just to move all the constructors, since there are multiple classes with constructors need to be relocated.

---
In reply to: [261314508](https://github.com/Azure/azure-relay-java/pull/26#discussion_r261314508) [](ancestors = 261314508)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: Thanks, that's nice to know!

---
In reply to: [261311954](https://github.com/Azure/azure-relay-java/pull/26#discussion_r261311954) [](ancestors = 261311954)

dlstucki: (The issue is someone could call setOnlineHandler(null) in-between the two getOnlineHandler calls in this snippet.)

---
In reply to: [261317780](https://github.com/Azure/azure-relay-java/pull/26#discussion_r261317780) [](ancestors = 261317780,261311954)

PR #27, opened 2019-03-01
Fix to properly reconnect controlConnection

bainian12345:


src/main/java/com/microsoft/azure/relay/TokenRenewer.java
```diff
@@ -62,6 +61,7 @@ void onRenewTimer() {
 	}
 
 	private void scheduleRenewTimer(SecurityToken token) {
+		this.renewTimer = new Timer();
```
bainian12345: >this.renewTimer = new Timer(); [](start = 2, length = 30)

If disconnect happens, TokenRenewer will close. Once the "renewTimer" object calls .cancel(), it will throw if we try to do anything with it. It can't just change to a new time like in C#, so I'm creating a new intance every time we schedule.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -152,7 +152,9 @@ boolean isOpen() {
 	 */
 	CompletableFuture<String> readTextAsync() {
 		return this.textQueue.dequeueAsync().thenApply(text -> {
-			RelayLogger.logEvent("receivedText", this, String.valueOf(text.length()));
+			if (text != null) {
```
dlstucki: >text [](start = 7, length = 4)

Does a null text ever happen? Maybe at close time? Do you want any trace then?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -766,7 +766,9 @@ public void close() {
 								}
 								return keepGoing;
 							}
-							this.listener.onCommandAsync(receivedMessage, webSocket);
+							if (receivedMessage != null) {
```
dlstucki: >receivedMessage [](start = 11, length = 15)

Under what circumstances does a null receiveMessage arrive? if it's null should keepGoing be true or false?

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
bainian12345: That's the reason why I put a null check. Text can be null after close and text.length() will give a null pointer.

---
In reply to: [261783196](https://github.com/Azure/azure-relay-java/pull/27#discussion_r261783196) [](ancestors = 261783196)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: Theoretically it should never happen, since the only possible cause is the control websocket closes. Just want to make sure it won't fall into the line below because it will cause exception.

---
In reply to: [261783674](https://github.com/Azure/azure-relay-java/pull/27#discussion_r261783674) [](ancestors = 261783674)

bainian12345: I guess keepGoing should be true by default, unless closeCalled is true, which should be caught in the block above.

---
In reply to: [261791257](https://github.com/Azure/azure-relay-java/pull/27#discussion_r261791257) [](ancestors = 261791257,261783674)

PR #28, opened 2019-03-01
Added more tests to handlers and reconnection for Listener

bainian12345:


src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
+			assertNull("There should not have been an exception when closing the listener normally.", ex);
+			offlineHandlerCalled.set(true);
+		});
+		
+		listener.openAsync(Duration.ofSeconds(15)).join();
+		assertTrue("Listener failed to open.", listener.isOnline());
+		assertTrue("Listener open handler was not called", onlineHandlerCalled.get());
+		
+		listener.close();
+		assertFalse("Listner should be closed", listener.isOnline());
```
dlstucki: >Listner [](start = 15, length = 7)

Nit: typo

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
 	
-	@AfterClass
-	public static void cleanup() {
-		if (listener != null) {
-			listener.close();
-		}
+	@Test
+	public void openAndCloseTest() {
+		HybridConnectionListener listener = new HybridConnectionListener(CONNECTION_URI, tokenProvider);
+		AtomicBoolean onlineHandlerCalled = new AtomicBoolean(false);
```
dlstucki: >AtomicBoolean [](start = 42, length = 13)

You could make these AtomicIntegers and then assert that they get called exactly one time (or however many it should be given the test).

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
+			return CompletableFutureUtil.delayAsync(Duration.ofMillis(waitMS), EXECUTOR).whenComplete(($void, ex) -> {
+				assertFalse("client connection should be closed implicitly when rendezvous connection closes", clientConnection.isOpen());
+			});
+		});
+		
+		CompletableFuture.allOf(clientClosing, rendezvousClosing).whenComplete(($void, ex) -> {
+			if (ex != null) {
+				fail(ex.getMessage());
+			}	
+			listener.close();
```
dlstucki: >listener.close(); [](start = 3, length = 17)

I'd strongly recommend putting the listener close inside a try/finally block. If a test case fails but leaves a listener connected to the same endpoint it could easily mess with other tests.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
+		
+		// Handler accepts the connection
+		listener.setAcceptHandler(context -> {
+			handlerExecuted.incrementAndGet();
+			return true;
 		});
-		assertTrue("Listener failed to accept connections from sender in webSocket mode.", checkSocketConnectionTask.join());
+		
+		client.createConnectionAsync().whenComplete((clientConnection, ex) -> {
+			assertTrue("clientConnection should have been accepted", clientConnection != null && clientConnection.isOpen());
```
dlstucki: >clientConnection [](start = 60, length = 16)

Who closes clientConnection?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
+		listener.openAsync().join();
+		
+		// Handler accepts the connection
+		listener.setAcceptHandler(context -> {
+			handlerExecuted.incrementAndGet();
+			return true;
 		});
-		assertTrue("Listener failed to accept connections from sender in webSocket mode.", checkSocketConnectionTask.join());
+		
+		client.createConnectionAsync().whenComplete((clientConnection, ex) -> {
```
dlstucki: >whenComplete [](start = 33, length = 12)

who waits on the CF<Void> that represents this createConnectionAsync+whenComplete? If nobody ever waits on it would a failedAssert still cause the test to fail? (I don't think it would, personally)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
+		client.createConnectionAsync().handle((clientConnection, ex) -> {
+			assertNull("clientConnection should not have been accepted", clientConnection);
+			assertNotNull("Rejection should have thrown.", ex);
+
+			RelayedHttpListenerResponse response = httpContext.get().getResponse();
+			assertEquals("The rejection response code is incorrect", HttpStatus.BAD_REQUEST_400, response.getStatusCode());
+			assertEquals("The rejection response description is incorrect", "Rejected by user code", response.getStatusDescription());
+			assertEquals("Both handlers should have been run", 2, handlerExecuted.get());
+			
+			listener.close();
```
dlstucki: >listener [](start = 3, length = 8)

If the test fails prior to this point does the listener stay connected?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -29,45 +33,135 @@
...
+			return false;
+		});
+		CompletableFuture<HybridConnectionChannel> acceptConnectionTask = listener.acceptConnectionAsync();
+		
+		client.createConnectionAsync().handle((clientConnection, ex) -> {
+			assertNull("clientConnection should not have been accepted", clientConnection);
+			assertNotNull("Rejection should have thrown.", ex);
+
+			RelayedHttpListenerResponse response = httpContext.get().getResponse();
+			assertEquals("The rejection response code is incorrect", HttpStatus.BAD_REQUEST_400, response.getStatusCode());
```
dlstucki: >HttpStatus.BAD_REQUEST_400, response.getStatusCode() [](start = 60, length = 52)

Is there any way to verify the error details that went over the wire instead of sneaking the internal object from the Listener to the Sender and checking its values?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -86,11 +181,39 @@ public void acceptHttpConnectionTest() throws IOException {
...
+		listener.close();
+	}
+	
+	@Test
+	public void listenerReconnectionTest() {
+		HybridConnectionListener listener = new HybridConnectionListener(CONNECTION_URI, tokenProvider);
+		AtomicBoolean handlerExecuted = new AtomicBoolean(false);
+		
+		listener.setConnectingHandler((ex) -> {
+			assertTrue("Exception should be ConnectionLostException.", ex != null && ex instanceof ConnectionLostException);
```
dlstucki: >assertTrue [](start = 3, length = 10)

I'm not sure asserting and throwing in an event handler is a good idea. Some would argue that's a violation of the implicit contract of an event handler.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -86,11 +181,39 @@ public void acceptHttpConnectionTest() throws IOException {
...
+			handlerExecuted.set(true);
+		});
+		
+		listener.openAsync(Duration.ofSeconds(15)).join();
+		ControlConnection controlConnection = listener.getControlConnection();
+		controlConnection.close(); // This bypasses listener.close(), simulating an unexptected close
```
dlstucki: >unexptected [](start = 78, length = 11)

nit: typo

dlstucki: >/ This bypasses listener.close( [](start = 30, length = 31)

this is great!

dlstucki: >close [](start = 20, length = 5)

Wait, I'm confused. What does calling controlConnection.close do? Will it reconnect after you call close?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -170,5 +293,6 @@ public void connectMultipleClientsTest() throws URISyntaxException {
 		
 		CompletableFuture.allOf(clientCloseFutures).join();
 		CompletableFuture.allOf(listenerCloseFutures).join();
+		listener.close();
```
dlstucki: >listener [](start = 2, length = 8)

I'd sleep easier at night if these were in finally blocks.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: If it's possible for the controlConnection to be in a state where it doesn't think it should reconnect but the outer HCL thinks things are in a state where it will reconnect could that lead to or hide potential issues?

---
In reply to: [261787118](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261787118) [](ancestors = 261787118)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: It only gets called once though, at least in this test.

---
In reply to: [261784456](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261784456) [](ancestors = 261784456)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: What do you mean? We rely heavily on the "closeCalled" flag from the outer Listener, so if we call close on the inner ControlConnection, we can bypass the closeCalled flag. As far as the outer listener is concerned, it's an unexpected close and will try to reconnect.

---
In reply to: [261788947](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261788947) [](ancestors = 261788947,261787118)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: don't think so, the websocket mode doesn't deal with the http upgrade response.

---
In reply to: [261786150](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261786150) [](ancestors = 261786150)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: should be closed implicitly by the listener side rendezvous in the line below.

---
In reply to: [261785389](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261785389) [](ancestors = 261785389)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: How do you know it doesn't accidentally get called 2 or 3 times???

---
In reply to: [261791621](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261791621) [](ancestors = 261791621,261784456)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: does the listener side of the connection block until the sender side of the connection also calls close?

---
In reply to: [262209295](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262209295) [](ancestors = 262209295,261785389)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: The outer listener never tries to reconnect.

---
In reply to: [261800522](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261800522) [](ancestors = 261800522,261788947,261787118)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: The current approach bothers me. This code is checking default assigned values on an object which I'm not sure it's really allowed to be holding on to at this point.

---
In reply to: [261800909](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261800909) [](ancestors = 261800909,261786150)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -10,64 +10,166 @@
...
 public class HybridConnectionListenerTest {
-	// Max # simultaneous client connections = # of cores - 1
-	// Because one thread need to be reserved for listener
 	private static final int MAX_CONNECTIONS_COUNT = Math.max(1, Runtime.getRuntime().availableProcessors() * 2);
-	private static HybridConnectionListener listener;
+	private static final AutoShutdownScheduledExecutor EXECUTOR = AutoShutdownScheduledExecutor.Create();
 	private static TokenProvider tokenProvider;
 	private static HybridConnectionClient client;
 	private static URI CONNECTION_URI;
+	private static HybridConnectionListener listener;
```
dlstucki: >static [](start = 9, length = 6)

does this need to be static?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -10,64 +10,166 @@
...
-			listener.close();
-		}
+	@Before
+	public void beforeEachTest() {
+		listener = new HybridConnectionListener(CONNECTION_URI, tokenProvider);
+	}
+	
+	@After
+	public void afterEachTest() {
+		listener.close();
```
dlstucki: >close [](start = 11, length = 5)

Is listener.close throws does the test case fail?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: Iteration 3 is wonderful!

---
In reply to: [262212380](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262212380) [](ancestors = 262212380,261800909,261786150)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: if listener.close() throws it will fail that test

---
In reply to: [262216035](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262216035) [](ancestors = 262216035)

bainian12345: this @After attribute acts as a final block for each test, so it behaves like part of each test.

---
In reply to: [262223458](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262223458) [](ancestors = 262223458,262216035)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: it doesn't have to be, but why not?

---
In reply to: [262215879](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262215879) [](ancestors = 262215879)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -586,9 +585,7 @@ public Throwable getLastError() {
 				return connectTask.thenCompose((webSocket) -> {
 					return this.sendAsyncLock.acquireThenCompose(duration, () -> {
 						CloseReason reason = new CloseReason(CloseCodes.NORMAL_CLOSURE, "Normal Closure");					
-						return webSocket.closeAsync(reason).whenComplete(($void, ex) -> {
-							this.onOffline(ex);
-						});
```
bainian12345: Offline handler was called twice. Removing this one here because the other one is located after the ReceivePump exits, which is executed even when the close is unexpected.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: No it does not block. Closing this connection implicitly is part of the previous test case, so if it doesn't happen it should be caught in the other test.

---
In reply to: [262210594](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262210594) [](ancestors = 262210594,262209295,261785389)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: Because if you instantiate two instances of this class and call beforeEachTest on both of them they should both use the same (last) listener.

---
In reply to: [262227695](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262227695) [](ancestors = 262227695,262215879)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: I think this is fixed now, right?

---
In reply to: [261785842](https://github.com/Azure/azure-relay-java/pull/28#discussion_r261785842) [](ancestors = 261785842)

bainian12345: it's fixed

---
In reply to: [262252931](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262252931) [](ancestors = 262252931,261785842)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: Created a new "closeCalled" flag for ControlConnection as well so it wouldn't reconnect

---
In reply to: [262211154](https://github.com/Azure/azure-relay-java/pull/28#discussion_r262211154) [](ancestors = 262211154,261800522,261788947,261787118)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -525,6 +525,7 @@ public Throwable getLastError() {
 			return lastError;
 		}
 		
+		// For test and debugging access only
 		CompletableFuture<ClientWebSocket> getConnectAsyncTask() {
```
dlstucki: >getConnectAsyncTask [](start = 37, length = 19)

Sweet! You already had a way to get at the internal guts and kill it!

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -577,6 +578,7 @@ public Throwable getLastError() {
 			
 			CompletableFuture<ClientWebSocket> connectTask;
 			synchronized (this.thisLock) {
+				this.closeCalled = true;
```
dlstucki: if you wanted this would be a nice place (inside the sync block) to add:

if (this.closeCalled) {
   return CompletableFuture.completed(null);
}

PR #29, opened 2019-03-06
Terminate running threads when ClientWebSocket closes

bainian12345:


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -323,6 +322,14 @@ public void onMessage(byte[] inputBytes, boolean isEnd) {
 	
 	@OnClose
 	public void onClose(Session session, CloseReason reason) {
+		CompletableFuture.runAsync(() -> {
+			try {
+				((LifeCycle) this.container).stop();
```
bainian12345: >((LifeCycle) this.container).stop(); [](start = 4, length = 36)

For some reason this must be put in a separate thread and not waited upon for it to work. If you ask me why, I'm not sure, but it does execute and do what it's supposed to do...

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -289,7 +289,6 @@ else if (data instanceof String){
 			} else {
 				this.session.close();
 			}
-			((LifeCycle) this.container).stop();
```
bainian12345: >((LifeCycle) this.container).stop(); [](start = 3, length = 36)

Moving this to the @OnClose block below so ClientWebSockets that close passively will terminate its running threads too.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
dlstucki: does it deadlock if called on the thread invoking onClose? That wouldn't surprise me.

---
In reply to: [263108491](https://github.com/Azure/azure-relay-java/pull/29#discussion_r263108491) [](ancestors = 263108491)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -203,8 +203,8 @@ public void listenerReconnectionTest() {
 		
 		listener.openAsync(Duration.ofSeconds(15)).join();
 		ControlConnection controlConnection = listener.getControlConnection();
-		controlConnection.getConnectAsyncTask().thenAccept(controlWebSocket -> {
-			controlWebSocket.closeAsync(); // This bypasses listener.close(), simulating an unexpected close
+		controlConnection.getConnectAsyncTask().thenCompose(controlWebSocket -> {
```
dlstucki: >thenCompose [](start = 42, length = 11)

good catch.

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -203,16 +205,17 @@ public void httpLargePOSTAndLargeResponseTest() throws IOException {
...
+		AtomicBoolean receivedReply = new AtomicBoolean(false);
 		
 		return client.createConnectionAsync().thenAccept((channel) -> {
 			channel.writeAsync(StringUtil.toBuffer(msgToSend)).join();
 			channel.readAsync().thenAccept((bytesReceived) -> {
 				String msgReceived = new String(bytesReceived.array());
-				receivedReply.complete(true);
+				receivedReply.set(true);
 				assertEquals("Websocket sender did not receive the expected reply.", msgExpected, msgReceived);
+				channel.closeAsync().join();
```
dlstucki: >join [](start = 25, length = 4)

instead of join here might it be better to change "thenAccept" to "thenCompose"? and `return channel.closeAsync();`?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -203,16 +205,17 @@ public void httpLargePOSTAndLargeResponseTest() throws IOException {
...
 			});
-		}).thenRun(() -> assertTrue("Did not receive message from websocket sender.", receivedReply.join()));
```
dlstucki: >join [](start = 94, length = 4)

CF<bool>.join was capable of waiting until it got signaled. Is there any chance the AtomicBoolean might not be set yet due to it executing async (and I think nobody waits for it)?

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
bainian12345: That's a likely cause. Are there better solutions than just simply putting it on a different thread?

---
In reply to: [263109139](https://github.com/Azure/azure-relay-java/pull/29#discussion_r263109139) [](ancestors = 263109139,263108491)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: this stage only executes after the previous stage where the AtomicBoolean is set, so it should be fine if the previous stage changes to .thenCompose instead. If we use CF.join() then it may just hang here if things actually go wrong.

---
In reply to: [263111405](https://github.com/Azure/azure-relay-java/pull/29#discussion_r263111405) [](ancestors = 263111405)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
dlstucki: This seems perfectly fine, or use executor.submit to use our HybridConnectionClient thread pool?

We have ActionItem.schedule but that's implemented with CF.runAsync anyhow!

---
In reply to: [263111543](https://github.com/Azure/azure-relay-java/pull/29#discussion_r263111543) [](ancestors = 263111543,263109139,263108491)

dlstucki: Actually ActionItem should probably be removed. It's only used by InputQueue. That class should use its own executor instead.

---
In reply to: [263112694](https://github.com/Azure/azure-relay-java/pull/29#discussion_r263112694) [](ancestors = 263112694,263111543,263109139,263108491)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: Ok, CF.join could certainly deadlock. Something with a timeout would have been better.

I don't see anyone waiting on the channel.readAsync task tree so I'm still not clear how the thenRun would definitely wait until channel.readAsync + thenAccept runs.

---
In reply to: [263112543](https://github.com/Azure/azure-relay-java/pull/29#discussion_r263112543) [](ancestors = 263112543,263111405)

PR #30, opened 2019-03-06
Cleanup unused InputQueue code and remove ActionItem

bainian12345:


src/main/java/com/microsoft/azure/relay/InputQueue.java
```diff
@@ -127,8 +118,8 @@ public void dispatch() {
 		}
 
 		if (outstandingReaders != null) {
-			ActionItem.schedule(s -> completeOutstandingReadersCallback(s), outstandingReaders);
-			
+			AtomicReference<Object> readersRef = new AtomicReference<>(outstandingReaders);
```
dlstucki: >AtomicReference<Object> readersRef = new AtomicReference<>(outstandingReaders); [](start = 3, length = 79)

I'm not sure I understand why this AtomicReference is needed.

dlstucki: What bad thing might happen if it were removed? (The C# pattern passed it as a parameter to avoid allocating something for the enclosure).

---
In reply to: [263167814](https://github.com/Azure/azure-relay-java/pull/30#discussion_r263167814) [](ancestors = 263167814)

bainian12345: The line below is a lambda function. It won't compile unless we use AtomicReference (Only final or local variables are allowed inside lambda functions). I don't know if you have run into enough of these cases but this has to be one of the most annoying things to deal with.

---
In reply to: [263168584](https://github.com/Azure/azure-relay-java/pull/30#discussion_r263168584) [](ancestors = 263168584,263167814)

dlstucki: Dang it! If executor allowed passing a state object ("Attachment"?) to an "Acceptor" then it would be easy:

executor.submit((readers) -> completeOutstandingReadersCallback(readers), outstandingReaders);

---
In reply to: [263170069](https://github.com/Azure/azure-relay-java/pull/30#discussion_r263170069) [](ancestors = 263170069,263168584,263167814)

dlstucki: Is this a reasonable work-around?
```
  if (outstandingReaders != null) {
   CompletableFuture<T>[] outstandingReadersRef = outstandingReaders;
   executor.submit(() -> completeOutstandingReadersCallback(outstandingReadersRef));
  }
```

---
In reply to: [263173921](https://github.com/Azure/azure-relay-java/pull/30#discussion_r263173921) [](ancestors = 263173921,263170069,263168584,263167814)

bainian12345: Yes, that would save an allocation. The "final" isn't needed, which is kind of stupid how we can get around this by just having extra variables pointing to the same thing...

---
In reply to: [263176314](https://github.com/Azure/azure-relay-java/pull/30#discussion_r263176314) [](ancestors = 263176314,263173921,263170069,263168584,263167814)

dlstucki: yeah, I edited and added the final as opposed to a comment about how the compiler wanted it to be effectively final. You pick.

---
In reply to: [263178892](https://github.com/Azure/azure-relay-java/pull/30#discussion_r263178892) [](ancestors = 263178892,263176314,263173921,263170069,263168584,263167814)

PR #31, opened 2019-03-08
Add new test to verify that the listener receives the correct path an…

bainian12345:


src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -191,6 +192,55 @@ public void requestHandlerTest() throws IOException {
...
+		listener.setRequestHandler((context) -> {
+			try {
+				assertNotNull("Listener should have received a valid http context.", context);
+				assertNotNull("Listener should have received a valid http request.", context.getRequest());
+				
+				URI requestUri = context.getRequest().getUri();
+				assertNotNull("Listener should have a URI from request", requestUri);
+				assertEquals(
+						"The path wasn't the expected value", 
+						"/bailiu-relay-hc/bailiu-relay-hc" + extraPath,
```
dlstucki: >bailiu-relay-hc/bailiu-relay-hc [](start = 8, length = 31)

This won't work for other people, right?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -191,6 +192,55 @@ public void requestHandlerTest() throws IOException {
...
+				assertEquals(
+						"The path wasn't the expected value", 
+						"/bailiu-relay-hc/bailiu-relay-hc" + extraPath, 
+						requestUri.getPath()
+				);
+				assertEquals("The query wasn't the expected value", extraQuery, requestUri.getQuery());
+				
+				requestReceived.complete(null);
+			} catch (Throwable ex) {
+				requestReceived.completeExceptionally(ex);
```
dlstucki: I love this. At first I wondered how asserts in this handler would get reported.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: I think it should be based on TestUtil.ENTITY_PATH

---
In reply to: [263873685](https://github.com/Azure/azure-relay-java/pull/31#discussion_r263873685) [](ancestors = 263873685)

bainian12345: Yes it should be TestUtil.ENTITY_PATH, although I find it weird why it's appended again.

---
In reply to: [263873685](https://github.com/Azure/azure-relay-java/pull/31#discussion_r263873685) [](ancestors = 263873685)

dlstucki: Is that a bug?

---
In reply to: [263976580](https://github.com/Azure/azure-relay-java/pull/31#discussion_r263976580) [](ancestors = 263976580,263873685)

PR #32, opened 2019-03-11
Moved all constructors to top of file and cleaned up some access modifiers

bainian12345:


src/main/java/com/microsoft/azure/relay/HybridConnectionClient.java
```diff
@@ -174,6 +128,52 @@ public HybridConnectionClient(String connectionString, String path) throws URISy
...
+	/**
+	 * The TokenProvider for authenticating this HybridConnection listener.
+	 */
+	private TokenProvider tokenProvider;
+
+	/**
+	 * The default timeout for connecting a HybridConnection. Default value is 70
+	 * seconds.
+	 */
+	private Duration operationTimeout;
```
dlstucki: I thought fields went above constructors which getters/setters went below constructors.

PR #33, opened 2019-03-11
Parse operation timeout as long instead of string

bainian12345:


src/main/java/com/microsoft/azure/relay/RelayConnectionStringBuilder.java
```diff
@@ -144,7 +144,7 @@ public String toString() {
 
 		if (this.operationTimeout != RelayConstants.DEFAULT_OPERATION_TIMEOUT) {
 			connectionStringBuilder.append(OPERATION_TIMEOUT_CONFIG_NAME + KEY_VALUE_SEPARATOR
-					+ this.operationTimeout.toString() + KEY_VALUE_PAIR_DELIMITER);
+					+ this.operationTimeout.getSeconds() + KEY_VALUE_PAIR_DELIMITER);
```
dlstucki: >getSeconds [](start = 29, length = 10)

Just to be certain, does this return a long/int? Just don't want it to ever have a fractional component.

src/main/java/com/microsoft/azure/relay/RelayConnectionStringBuilder.java
```diff
@@ -244,13 +244,10 @@ private void parseConnectionString(String connectionString) {
...
-					this.operationTimeout = timeValue;
-				} catch (DateTimeParseException e) {
-					throw new DateTimeParseException(value + " cannot be parsed into a valid duration.",
-							e.getParsedString(), e.getErrorIndex());
+					this.operationTimeout = Duration.ofSeconds(Long.parseLong(value));
+				} catch (NumberFormatException e) {
+					throw RelayLogger.throwingException(e, this);
 				}
 			} else {
 				throw new IllegalArgumentException("the following is not a valid field for connection string: " + key);
```
dlstucki: >throw [](start = 4, length = 5)

[Unrelated to these changes] While I'm here do we want to have RelayLogger.throwingException wrapper around this?

PR #35, opened 2019-03-15
allow HybridConnectClient to connect with custom headers

bainian12345:


src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -71,6 +75,49 @@ public void openAndCloseTest() {
...
 	}
 	
+	@Test
+	public void customHeadersTest() {
+		String key = "GoodHeader";
+		String value = "GoodValue";
+		Map<String, List<String>> customHeaders = new HashMap<String, List<String>>();
+		customHeaders.put(key, Arrays.asList(new String[]{value}));
+		
+		CompletableFuture<Void> requestReceived = new CompletableFuture<Void>();
```
dlstucki: >requestReceived [](start = 26, length = 15)

Doesn't someone need to wait on this in case an exception was thrown?

src/main/java/com/microsoft/azure/relay/HybridConnectionClient.java
```diff
@@ -190,14 +200,17 @@ public String toString() {
...
+		}
 		ClientEndpointConfig config = ClientEndpointConfig.Builder.create().configurator(configurator).build();
 
 		try {
 			URI uri = HybridConnectionUtil.buildUri(this.address.getHost(), this.address.getPort(),
 					this.address.getPath(), this.address.getQuery(), HybridConnectionConstants.Actions.CONNECT,
 					trackingContext.getTrackingId());
 			WebSocketChannel channel = new WebSocketChannel(trackingContext, EXECUTOR);
-			return channel.getWebSocket().connectAsync(uri, this.operationTimeout, config).thenApply(result -> channel);
+			return channel.getWebSocket().connectAsync(uri, this.operationTimeout, config).thenApply($void -> channel);
```
dlstucki: >$void [](start = 92, length = 5)

I was toying around when I came up with naming these throw-away parameters `$void`. Seems like it might have caught on?

bainian12345: I liked it, seems clear that it's a null result

---
In reply to: [266164513](https://github.com/Azure/azure-relay-java/pull/35#discussion_r266164513) [](ancestors = 266164513)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: Ooops, deleted the wait afterwards by accident, will add it.

---
In reply to: [266164217](https://github.com/Azure/azure-relay-java/pull/35#discussion_r266164217) [](ancestors = 266164217)

PR #36, opened 2019-03-20
Allow TokenRenew command to be sent + TokenProvider cleanup

bainian12345:


src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -615,7 +615,14 @@ public Throwable getLastError() {
...
 				.thenCompose(webSocket -> {
 					return sendAsyncLock.acquireThenCompose(timeout, () -> {
-						String json = command.getResponse().toJsonString();
+						String json = null;
+						if (command.getResponse() != null) {
+							json = command.getResponse().toJsonString();
+						}
+						if (command.getRenewToken() != null) {
+							json = command.getRenewToken().toJsonString();
+						}
```
dlstucki: Logic like this could be made into a ListenerCommand.toJsonString() method. Or you could get rid of ListenerCommand all together and have each command implement an interface with a method like toJsonString() on it.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -615,7 +615,14 @@ public Throwable getLastError() {
 			return this.ensureConnectTask(timeout)
 				.thenCompose(webSocket -> {
 					return sendAsyncLock.acquireThenCompose(timeout, () -> {
-						String json = command.getResponse().toJsonString();
+						String json = null;
+						if (command.getResponse() != null) {
+							json = command.getResponse().toJsonString();
+						}
+						if (command.getRenewToken() != null) {
```
dlstucki: >if [](start = 6, length = 2)

should this be an `else if` instead? What if none of the if statements match? (All good reasons to encapsulate this inside ListenerCommand.toJsonString(), possibly making ListenerCommand into an interface with toJsonString method).

dlstucki:
> 				throw new IllegalArgumentException("invalid commnd type received.");

would you mind fixing this typo while we're here?

---
Refers to: src/main/java/com/microsoft/azure/relay/ListenerCommand.java:51 in 94d3951. [](commit_id = 94d39511307c6c437548609bf4b60019d7103a5c, deletion_comment = False)

dlstucki:
> class ListenerCommand {

How intrusive would it be to make ListenerCommand an abstract class,, have Accept, RenewToken, Request, etc. extend ListenerCommand and implement toJsonString.

abstract class ListenerCommand {
    public **static** ListenerCommand fromJsonString(String json);

    public abstract String toJsonString();
}

class RequestCommand extends ListenerCommand {
    toJsonString() { ... }
}


---
Refers to: src/main/java/com/microsoft/azure/relay/ListenerCommand.java:11 in 94d3951. [](commit_id = 94d39511307c6c437548609bf4b60019d7103a5c, deletion_comment = False)

src/main/java/com/microsoft/azure/relay/TokenProvider.java
```diff
@@ -50,19 +50,20 @@ public static TokenProvider createSharedAccessSignatureTokenProvider(String keyN
...
-		TimeoutHelper.throwIfNegativeArgument(validFor, "validFor");
 		CompletableFuture<SecurityToken> future = new CompletableFuture<SecurityToken>();
-		try {
-			urlString = normalizeAudience(urlString);
-			future = this.onGetTokenAsync(urlString, validFor);
-		}
-		catch (URISyntaxException e) {
+		if (StringUtil.isNullOrEmpty(urlString)) {
+			IllegalArgumentException e = new IllegalArgumentException("Null url provided for security token");
+			RelayLogger.handledExceptionAsWarning(e, this);
```
dlstucki: >handledExceptionAsWarning [](start = 15, length = 25)

handledException or throwingException?

src/main/java/com/microsoft/azure/relay/TokenProvider.java
```diff
@@ -50,19 +50,20 @@ public static TokenProvider createSharedAccessSignatureTokenProvider(String keyN
...
 			future.completeExceptionally(e);
+		} else {
+			try {
+				TimeoutHelper.throwIfNegativeArgument(validFor, "validFor");
```
dlstucki: >throwIfNegativeArgument [](start = 18, length = 23)

This parameter validation actually throws and is handled by the catch block. With these changes the parameter validations use two different approaches to this. (Also one traces, the other does not).

src/main/java/com/microsoft/azure/relay/TokenProvider.java
bainian12345: Should a null token string be a fatal exception?

---
In reply to: [267897192](https://github.com/Azure/azure-relay-java/pull/36#discussion_r267897192) [](ancestors = 267897192)

dlstucki: I'm not sure I understand. I think of "handledException" as "someone else threw an exception, I'm catching it, reacting to it, and not re-throwing anything, upstream callers won't see this exception".
I think of "throwingException" as "something went wrong, I'm throwing an exception so upstream callers know something is wrong".

---
In reply to: [267922252](https://github.com/Azure/azure-relay-java/pull/36#discussion_r267922252) [](ancestors = 267922252,267897192)

bainian12345:
> class ListenerCommand {

Well do in a later commit

---
In reply to: [475346273](https://github.com/Azure/azure-relay-java/pull/36#issuecomment-475346273) [](ancestors = 475346273)

---
Refers to: src/main/java/com/microsoft/azure/relay/ListenerCommand.java:11 in 94d3951. [](commit_id = 94d39511307c6c437548609bf4b60019d7103a5c, deletion_comment = False)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: I'm hoping that with the ListenerCommand changes this if/else-if will turn into a single line calling a method which hides any complexity like this.

---
In reply to: [267891127](https://github.com/Azure/azure-relay-java/pull/36#discussion_r267891127) [](ancestors = 267891127)

PR #38, opened 2019-04-12
Fix response stream.close

bainian12345:
Fix null pointer exception that may occur if HybridHttpConnection did not have a rendezvous websocket.


src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -425,7 +425,11 @@ public String toString() {
 
 				return sendTask.thenCompose((result) -> {
 					this.closed = true;
-					return closeRendezvousAsync();
+					if (this.connection.rendezvousWebSocket != null) {
+						return closeRendezvousAsync();
+					} else {
+						return CompletableFuture.completedFuture(null);
+					}
```
dlstucki: This if check could alternatively be done first thing in closeRendezvousAsync.

PR #39, opened 2019-04-15
Fix a few potential exceptions with Http mode

bainian12345:
- Use ByteArrayOutoutStream instead of ByteBuffer
- Fix a few potential null pointers
- Fix potential race with onWriteBufferFlushTimer

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -1,9 +1,9 @@
 package com.microsoft.azure.relay;
 
+import java.io.ByteArrayOutputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -381,8 +381,8 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 		return this.cachedString;
     }
 
-	CompletableFuture<Void> sendControlCommandAndStreamAsync(ListenerCommand command, ByteBuffer buffer, Duration timeout) {
-		return this.controlConnection.sendCommandAndStreamAsync(command, buffer, timeout);
+	CompletableFuture<Void> sendControlCommandAndStreamAsync(ListenerCommand command, ByteArrayOutputStream writeBufferStream, Duration timeout) {
+		return this.controlConnection.sendCommandAndStreamAsync(command, writeBufferStream, timeout);
 	}
 
 	void throwIfDisposed() throws RelayException {
@@ -605,12 +605,12 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 		 * if it exists
 		 * 
 		 * @param command The Listener command to be sent
-		 * @param buffer  The message body to be sent, null if it doesn't exist
+		 * @param writeBufferStream The message body to be sent, null if it doesn't exist
 		 * @param timeout The timeout to send within
 		 * @return Returns a completableFuture which completes when the command and
 		 *         stream are finished sending
 		 */
-		private CompletableFuture<Void> sendCommandAndStreamAsync(ListenerCommand command, ByteBuffer buffer, Duration timeout) {
+		private CompletableFuture<Void> sendCommandAndStreamAsync(ListenerCommand command, ByteArrayOutputStream writeBufferStream, Duration timeout) {
 
 			return this.ensureConnectTask(timeout)
 				.thenCompose(webSocket -> {
@@ -627,8 +627,8 @@ public class HybridConnectionListener implements RelayTraceSource, AutoCloseable
 							RelayLogger.logEvent("sendCommand", this, json);
 							return webSocket.writeAsync(json, timeout, WriteMode.TEXT)
 								.thenCompose($void -> {
-									if (buffer != null) {
-										return webSocket.writeAsync(buffer.array());
+									if (writeBufferStream != null) {
+										return webSocket.writeAsync(writeBufferStream.toByteArray());
 									} else {
 										return CompletableFuture.completedFuture(null);
 									}
--- a/src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
@@ -1,10 +1,11 @@
 package com.microsoft.azure.relay;
 
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Timer;
@@ -201,12 +202,12 @@ class HybridHttpConnection implements RelayTraceSource {
 	}
 
 	private CompletableFuture<Void> sendResponseAsync(ListenerCommand.ResponseCommand responseCommand,
-			ByteBuffer responseBodyBuffer, Duration timeout) throws CompletionException {
+			ByteArrayOutputStream writeBufferStream, Duration timeout) {
 		if (this.rendezvousWebSocket == null) {
 			RelayLogger.logEvent("httpSendResponse", this, "control", String.valueOf(responseCommand.getStatusCode()));
 			ListenerCommand listenerCommand = new ListenerCommand(null);
 			listenerCommand.setResponse(responseCommand);
-			return this.listener.sendControlCommandAndStreamAsync(listenerCommand, responseBodyBuffer, timeout)
+			return this.listener.sendControlCommandAndStreamAsync(listenerCommand, writeBufferStream, timeout)
 					.thenRun(() -> RelayLogger.logEvent("httpSendResponseFinished", this, "control", String.valueOf(responseCommand.getStatusCode())));
 		} else {
 			RelayLogger.logEvent("httpSendResponse", this, "rendezvous", String.valueOf(responseCommand.getStatusCode()));
@@ -221,10 +222,10 @@ class HybridHttpConnection implements RelayTraceSource {
 					.thenAccept(bytesWritten -> {
 						RelayLogger.logEvent("httpSendResponseFinished", this, "rendezvous", String.valueOf(responseCommand.getStatusCode()));
 					});
-			if (responseCommand.hasBody() && responseBodyBuffer != null) {
+			if (responseCommand.hasBody() && writeBufferStream != null) {
 				return sendCommandTask.thenCompose((result) -> {
-					int bytesToWrite = responseBodyBuffer.remaining();
-					return this.rendezvousWebSocket.writeAsync(responseBodyBuffer.array(), timeout).thenAccept(nullResult -> {
+					int bytesToWrite = writeBufferStream.size();
+					return this.rendezvousWebSocket.writeAsync(writeBufferStream.toByteArray(), timeout).thenAccept(nullResult -> {
 						RelayLogger.logEvent("httpSendingBytes", this, String.valueOf(bytesToWrite));
 					});
 				});
@@ -233,10 +234,12 @@ class HybridHttpConnection implements RelayTraceSource {
 		}
 	}
 
-	private CompletableFuture<Void> sendBytesOverRendezvousAsync(ByteBuffer buffer, Duration timeout) {
-		int bytesToWrite = buffer.remaining();
-		return this.rendezvousWebSocket.writeAsync(buffer, timeout).thenAccept(nullResult -> {
-			RelayLogger.logEvent("httpSendingBytes", this, String.valueOf(bytesToWrite));
+	private CompletableFuture<Void> sendBytesOverRendezvousAsync(ByteArrayOutputStream writeBufferStream, Duration timeout) {
+		if (writeBufferStream == null) {
+			return CompletableFuture.completedFuture(null);
+		}
+		return this.rendezvousWebSocket.writeAsync(writeBufferStream.toByteArray(), timeout).thenAccept(nullResult -> {
+			RelayLogger.logEvent("httpSendingBytes", this, String.valueOf(writeBufferStream.size()));
 		});
 	}
 
@@ -279,7 +282,7 @@ class HybridHttpConnection implements RelayTraceSource {
 		private final RelayedHttpListenerContext context;
 		private final AsyncLock asyncLock;
 		private boolean closed;
-		private ByteBuffer writeBufferStream;
+		private ByteArrayOutputStream writeBufferStream;
 		private Timer writeBufferFlushTimer;
 		private boolean responseCommandSent;
 		private final TrackingContext trackingContext;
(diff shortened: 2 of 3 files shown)


bainian12345:
> 										onWriteBufferFlushTimer();

This is not joined, so it may cause races. Either we should join here or make the method not async

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java:376 in 2214480. [](commit_id = 22144801fdfa9cc764366b315b3428ba5c813bff, deletion_comment = False)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -358,14 +361,13 @@ public void write(String text) {
...
 						flushReason = FlushReason.RENDEZVOUS_EXISTS;
 					} else {
-						int bufferedCount = this.writeBufferStream != null ? this.writeBufferStream.position() : 0;
+						int bufferedCount = this.writeBufferStream != null ? this.writeBufferStream.size() : 0;
 						if (count + bufferedCount <= MAX_CONTROL_CONNECTION_BODY_SIZE) {
 
 							// There's still a chance we might be able to respond over the control
 							// connection, accumulate bytes
 							if (this.writeBufferStream == null) {
-								int initialStreamSize = Math.min(count, MAX_CONTROL_CONNECTION_BODY_SIZE);
```
bainian12345: >int initialStreamSize = Math.min(count, MAX_CONTROL_CONNECTION_BODY_SIZE); [](start = 8, length = 74)

Don't think we need to do this min check, "count" won't be greater than MAX_CONTROL_CONNECTION_BODY_SIZE if we got here

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -358,14 +361,13 @@ public void write(String text) {
...
-								this.writeBufferStream = ByteBuffer.allocate(initialStreamSize);
+								this.writeBufferStream = new ByteArrayOutputStream(count);
```
bainian12345: >new ByteArrayOutputStream(count) [](start = 33, length = 32)

ByteArrayOutputStream can auto increase capacity, unlike ByteBuffer. Used to cause exception here with ByteBuffer if multiple writes were called that were under 64kb combined.

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
```diff
@@ -82,7 +82,9 @@ public RelayedHttpListenerContext getContext() {
 	 * RelayedHttpListenerResponse instance.
 	 */
 	public CompletableFuture<Void> closeAsync() {
-		return this.outputStream.closeAsync().whenComplete(($void, ex) -> this.disposed = true);
```
bainian12345: >this.outputStream [](start = 9, length = 17)

Potentially this.outputStream == null

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -381,8 +381,8 @@ public String toString() {
 		return this.cachedString;
     }
 
-	CompletableFuture<Void> sendControlCommandAndStreamAsync(ListenerCommand command, ByteBuffer buffer, Duration timeout) {
-		return this.controlConnection.sendCommandAndStreamAsync(command, buffer, timeout);
+	CompletableFuture<Void> sendControlCommandAndStreamAsync(ListenerCommand command, ByteArrayOutputStream writeBufferStream, Duration timeout) {
```
dlstucki: >ByteArrayOutputStream writeBufferStream, [](start = 83, length = 40)

Does changing from ByteBuffer to ByteArrayOutputStream eliminate potential exceptions in some way? How?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -627,8 +627,8 @@ else if (command.getRenewToken() != null) {
 							RelayLogger.logEvent("sendCommand", this, json);
 							return webSocket.writeAsync(json, timeout, WriteMode.TEXT)
 								.thenCompose($void -> {
-									if (buffer != null) {
-										return webSocket.writeAsync(buffer.array());
+									if (writeBufferStream != null) {
+										return webSocket.writeAsync(writeBufferStream.toByteArray());
```
dlstucki: >toByteArray [](start = 56, length = 11)

Ugh, I don't think this is any different than how it was before but any time we go from a ByteBuffer or Stream to an array of the exact right size we're basically making it impossible to pool those buffers.

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -627,8 +627,8 @@ else if (command.getRenewToken() != null) {
 							RelayLogger.logEvent("sendCommand", this, json);
 							return webSocket.writeAsync(json, timeout, WriteMode.TEXT)
 								.thenCompose($void -> {
-									if (buffer != null) {
-										return webSocket.writeAsync(buffer.array());
```
dlstucki: >array [](start = 45, length = 5)

Was the problem that the ByteBuffer.array might include extra data at the end?

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -233,10 +234,12 @@ void invokeRequestHandler(RequestCommandAndStream requestAndStream) {
...
-	private CompletableFuture<Void> sendBytesOverRendezvousAsync(ByteBuffer buffer, Duration timeout) {
-		int bytesToWrite = buffer.remaining();
-		return this.rendezvousWebSocket.writeAsync(buffer, timeout).thenAccept(nullResult -> {
-			RelayLogger.logEvent("httpSendingBytes", this, String.valueOf(bytesToWrite));
+	private CompletableFuture<Void> sendBytesOverRendezvousAsync(ByteArrayOutputStream writeBufferStream, Duration timeout) {
+		if (writeBufferStream == null) {
+			return CompletableFuture.completedFuture(null);
+		}
+		return this.rendezvousWebSocket.writeAsync(writeBufferStream.toByteArray(), timeout).thenAccept(nullResult -> {
+			RelayLogger.logEvent("httpSendingBytes", this, String.valueOf(writeBufferStream.size()));
```
dlstucki: >size [](start = 83, length = 4)

does size() still return the correct length after toByteArray was called?

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -415,12 +422,10 @@ public String toString() {
 					ListenerCommand.ResponseCommand responseCommand = createResponseCommand(this.context);
 					if (this.writeBufferStream != null) {
 						responseCommand.setBody(true);
-						this.writeBufferStream.position(0);
 					}
-
+					
 					// Don't force any rendezvous now
 					sendTask = this.connection.sendResponseAsync(responseCommand, this.writeBufferStream, this.writeTimeout);
-					this.responseCommandSent = true;
```
dlstucki: >responseCommandSent [](start = 10, length = 19)

This is now getting set later, is that intentional?

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -429,16 +434,20 @@ public String toString() {
...
 			});
 		}
 
-		CompletableFuture<Void> onWriteBufferFlushTimer() {
-			return this.asyncLock.acquireThenCompose(this.writeTimeout, () -> {
+		void onWriteBufferFlushTimer() {
+			this.asyncLock.acquireThenCompose(this.writeTimeout, () -> {
 				return this.flushCoreAsync(FlushReason.TIMER, this.writeTimeout);
-			});
+			}).join();
```
dlstucki: >).join(); [](start = 4, length = 9)

I need to go right now but this scares me. Can join be avoided?

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
bainian12345: Just putting it in the callback after the send completes.

---
In reply to: [276024545](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276024545) [](ancestors = 276024545)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
bainian12345: size() returns the number of effective bytes in the underlying byte[], so it will still reference to the same thing and return the same result.

---
In reply to: [276023732](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276023732) [](ancestors = 276023732)

dlstucki:
Is it still inside the sendAsyncLock at that later point? Should it be set to true even in case of send exception?

________________________________

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: There is nothing different here, just changing the signature of the function for convenience sake.

**writeAsync()** only takes byte[] or ByteBuffer, so I don't think there is a better way. (I recall trying their OutputStream way but it just did not work, can try again)

---
In reply to: [276023248](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276023248) [](ancestors = 276023248)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: The switch to ByteArrayOutputStream is mostly because it's self-resizable. If the user writes multiple times, it can easily lead to exceptions if the ByteBuffer wasn't set to the right size in the beginning.

---
In reply to: [276023404](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276023404) [](ancestors = 276023404)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: For future consideration one could subclass ByteArrayOutputStream in order to provide access to the underlying byte[] without copying into a new one, and then create a ByteBuffer using that array and the size() for passing into webSocket.writeAsync. This would avoid allocating a new array and having to copy all the data.

---
In reply to: [276041048](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276041048) [](ancestors = 276041048,276023248)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
dlstucki: A comment that this makes it closer to the C# implementation (if you had compared them yourself) might have saved me some time.

---
In reply to: [276036735](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276036735) [](ancestors = 276036735,276024545)

dlstucki:
> 										onWriteBufferFlushTimer();

Can you describe the race which requires joining or not being async?

---
In reply to: [483437016](https://github.com/Azure/azure-relay-java/pull/39#issuecomment-483437016) [](ancestors = 483437016)

---
Refers to: src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java:376 in 2214480. [](commit_id = 22144801fdfa9cc764366b315b3428ba5c813bff, deletion_comment = False)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
dlstucki: Can you describe the race which requires joining or not being async? The only thing I can think of would be if the async method hasn't completed yet and the timer fired again but I see how that can happen.

---
In reply to: [276024831](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276024831) [](ancestors = 276024831)

bainian12345: It could be avoided, it's just that nobody checks on it for any exceptions. I was thinking that this join() shouldn't matter as much since this is on its own separate thread.

---
In reply to: [276024831](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276024831) [](ancestors = 276024831)

bainian12345: With nothing joining or waiting here, the lock area exits after this.flushCoreAsync is called, not after it finishes executing. So isn't it possible that closeAsync() is called while flushCoreAsync() is still executing?

Maybe joining on the this.flushCoreAsync() inside is more appropriate?

---
In reply to: [276314775](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276314775) [](ancestors = 276314775,276024831)

bainian12345: Removing the join(). It turns out to be some other issue which will be fixed later.

---
In reply to: [276314775](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276314775) [](ancestors = 276314775,276024831)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: It can just use ByteBuffer like before, just need to manually allocate a new, larger ByteBuffer when needed (ByteArrayOutputStream needs to resize anyways).

---
In reply to: [276301369](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276301369) [](ancestors = 276301369,276041048,276023248)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
dlstucki: It's pretty common for Timers to be implemented where thousands of timers get fired using the same thread (since creating then destroying a thread is historically relatively expensive).

If there are hundreds of other timers which need to be signaled while this async operation is in progress I don't want those timers to be delayed.

What if one of those timers was attempting to cancel this async operation because it was taking too long? That would be a form of a deadlock.

---
In reply to: [276352552](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276352552) [](ancestors = 276352552,276024831)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
dlstucki: It appears that ByteArrayOutputStream is probably the correct thing to use since it will allocate a new array and copy all the existing data into the new array. It's better to let that built in class do it then to reinvent that wheel with our own code.

---
In reply to: [275564416](https://github.com/Azure/azure-relay-java/pull/39#discussion_r275564416) [](ancestors = 275564416)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
dlstucki: Good catch.

---
In reply to: [275563868](https://github.com/Azure/azure-relay-java/pull/39#discussion_r275563868) [](ancestors = 275563868)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
dlstucki: I was anticipating seeing another iteration on this. Is that coming?

---
In reply to: [276738473](https://github.com/Azure/azure-relay-java/pull/39#discussion_r276738473) [](ancestors = 276738473,276352552,276024831)

bainian12345:
Abandoned

PR #40, opened 2019-04-18
Allow adding custom response header

bainian12345:
Allow custom response header to be added

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/ListenerCommand.java
+++ b/src/main/java/com/microsoft/azure/relay/ListenerCommand.java
@@ -338,6 +338,15 @@ class ListenerCommand {
 				fields.add("\"statusDescription\":\"" + this.statusDescription + "\"");
 			}
 
+			if (this.responseHeaders != null && !this.responseHeaders.isEmpty()) {
+				StringBuilder headerBuilder = new StringBuilder("\"responseHeaders\":{");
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/azure/relay/ListenerCommand.java
```diff
@@ -338,6 +338,15 @@ String toJsonString() {
 				fields.add("\"statusDescription\":\"" + this.statusDescription + "\"");
 			}
 
+			if (this.responseHeaders != null && !this.responseHeaders.isEmpty()) {
```
dlstucki: >if [](start = 3, length = 2)

Will you add a unit test that verifies this works?

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -237,6 +240,7 @@ public void requestHandlerTest() throws IOException {
 		conn.setRequestProperty("ServiceBusAuthorization", tokenString);
 
 		assertEquals("Response did not have the expected response code.", status, conn.getResponseCode());
+		assertEquals("Response did not container the expected header", headerVal, conn.getHeaderField(headerKey));
```
bainian12345: >er [](start = 40, length = 2)

will correct the typo

PR #41, opened 2019-04-18
Disable modification to the response status, description, and header after writing to its output stream.

bainian12345:
- Disable modification to the response status, description, and header after writing to its output stream.
- Create method to assert throw CompletableFuture result


dlstucki:
Can you review mine?

https://github.com/Azure/azure-relay-dotnet/pull/105

bainian12345:
Sure, I thought it was done already. I swear I even saw your change getting pushed in Jenkins.

---
In reply to: [484712075](https://github.com/Azure/azure-relay-java/pull/41#issuecomment-484712075) [](ancestors = 484712075)

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
```diff
@@ -6,20 +6,20 @@
 import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.jetty.http.HttpStatus;
-import org.eclipse.jetty.io.RuntimeIOException;
 
 public class RelayedHttpListenerResponse {
 	private final Map<String, String> headers;
 	private final RelayedHttpListenerContext context;
 	private boolean disposed;
+	private boolean readonly = false;
```
dlstucki: > = false [](start = 25, length = 8)

Does it build/compile/save/whatever without warning if the " = false" is omitted here?

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
```diff
@@ -82,7 +83,8 @@ public RelayedHttpListenerContext getContext() {
 	 * RelayedHttpListenerResponse instance.
 	 */
 	public CompletableFuture<Void> closeAsync() {
-		return this.outputStream.closeAsync().whenComplete(($void, ex) -> this.disposed = true);
+		CompletableFuture<Void> closeTask = (this.outputStream != null) ? this.outputStream.closeAsync() : CompletableFuture.completedFuture(null);
```
dlstucki: >CompletableFuture<Void> closeTask = (this.outputStream != null) ? this.outputStream.closeAsync() : CompletableFuture.completedFuture(null); [](start = 2, length = 139)

This is personal preference but for me this pretty much crosses my threshold to use an actual 'if' statement.

Just saying, your choice to "Won't Fix" if you want.

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
```diff
@@ -93,9 +95,48 @@ public void close() {
...
+		String msg = null;
 		if (this.disposed) {
-			throw new RuntimeIOException("this connection has already been closed.");
+			msg = "The HTTP status code, status description, and headers cannot be modified after the output stream has already been closed.";
+		} else if (this.readonly) {
+			msg = "The HTTP status code, status description, and headers cannot be modified after writing to the output.";
+		}
+		
+		if (msg != null) {
+			throw RelayLogger.throwingException(new IllegalStateException(msg), this.context);
```
dlstucki: >IllegalStateException [](start = 43, length = 21)

Just confirming, this the the correct exception type in both cases?

src/test/java/com/microsoft/azure/relay/Assertions.java
```diff
@@ -30,4 +32,14 @@ private Assertions() {
 		String message = "Expected exception of type '" + expectedType + "' to be thrown.";
 		throw new AssertionFailedError(message);
 	}
+	
+	public static <T extends Throwable> T assertCFThrows(Class<T> expectedType, CompletableFuture<?> cf) {
```
dlstucki: >assertCFThrows [](start = 39, length = 14)

Cool.

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -402,4 +394,89 @@ public void connectMultipleClientsTest() throws URISyntaxException {
 		CompletableFuture.allOf(clientCloseFutures).join();
 		CompletableFuture.allOf(listenerCloseFutures).join();
 	}
+	
+	@Test
+	public void closedOrReadonlyResponseTest() throws IOException {
```
dlstucki: >closedOrReadonlyResponseTest [](start = 13, length = 28)

This is a fantastic unit test IMO!

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
bainian12345: The value need to initialized in order to compile

---
In reply to: [277057009](https://github.com/Azure/azure-relay-java/pull/41#discussion_r277057009) [](ancestors = 277057009)

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
bainian12345: How do you define "correct"? To me "IllegalStateException" makes sense for both scenarios, let me know if you feel otherwise.

---
In reply to: [277057844](https://github.com/Azure/azure-relay-java/pull/41#discussion_r277057844) [](ancestors = 277057844)

dlstucki: In .NET this code throws InvalidOperationException if the operation cannot be performed at this time but the object is still viable for certain other operations. ObjectDisposedException if Close/Dispose has been called on the object.

I just didn't know if Java has any similar patterns like .NET does.

---
In reply to: [277068116](https://github.com/Azure/azure-relay-java/pull/41#discussion_r277068116) [](ancestors = 277068116,277057844)

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
dlstucki: I don't understand why this private boolean is different than `disposed`. The coding convention isn't consistent between lines 14 and 15.

---
In reply to: [277067283](https://github.com/Azure/azure-relay-java/pull/41#discussion_r277067283) [](ancestors = 277067283,277057009)

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
bainian12345: Using IOException for "disposed" because OutputStream throws it when attempting to write after close.

Using IllegalStateException because it's probably the closest to InvalidOperationException in C#: [stackoverflow answer](https://stackoverflow.com/questions/3165701/is-there-a-java-equivalent-of-cs-invalidoperationexception/3165733)

---
In reply to: [277070605](https://github.com/Azure/azure-relay-java/pull/41#discussion_r277070605) [](ancestors = 277070605,277068116,277057844)

src/main/java/com/microsoft/azure/relay/RelayedHttpListenerResponse.java
bainian12345: Didn't see this comment before I pushed...will change it

---
In reply to: [277072160](https://github.com/Azure/azure-relay-java/pull/41#discussion_r277072160) [](ancestors = 277072160,277067283,277057009)

PR #42, opened 2019-04-19
Use ByteArrayOutputStream instead of ByteBuffer

bainian12345:


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -261,6 +263,24 @@ else if (data instanceof String){
 		}
 	}
 	
+	/**
+	 * Write contents from provided ByteArrayOutputStream to this web socket's OutputStream
+	 * @param src The source ByteArrayOutputStream that contains the contents to be written
+	 * @param timeout Timeout to complete the write operation within
+	 * @return A CompletableFuture that completes when the output stream finishes write and close.
+	 */
+	public CompletableFuture<Void> writeFromByteArrayStream(ByteArrayOutputStream src, Duration timeout) {
```
dlstucki: >writeFromByteArrayStream [](start = 32, length = 24)

Async suffix?

dlstucki: >ByteArrayOutputStream [](start = 57, length = 21)

ByteArrayOutputStream sure seems a lot harder to use than .NET's MemoryStream.

PR #43, opened 2019-04-23
Use fix-sized 64kb ByteBuffer

bainian12345:
- Use fix-sized 64kb ByteBuffer for writing to HTTP response
- Fix failing unit test by correcting expected exception


src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -417,7 +417,7 @@ public void closedOrReadonlyResponseTest() throws IOException {
 				response.getHeaders().put(headerKey, headerVal);
 				modifiedAfterClose.complete(null);
 			} catch (Exception e) {
-				modifiedAfterClose.completeExceptionally(e);
+				modifiedAfterClose.completeExceptionally(e.getCause());
```
bainian12345: >e.getCause() [](start = 45, length = 12)

The exception was changed to IOException in the last check-in and it's wrapped by RuntimeException now

PR #45, opened 2019-06-06
Fixed bugs that prevented consecutive writes to HTTP response from working properly

bainian12345:
- Added lock to ClientWebSocket to prevent exception from the websocket API when multiple sends are attempted simultaneously.

- Added trimming to HTTP response buffer to ensure that empty bytes do not get sent.

- Added tests to ensure consecutive writes to HTTP response is working properly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>com.microsoft.azure</groupId>
 	<artifactId>azure-relay</artifactId>
-	<version>0.0.2-PREVIEW</version>
+	<version>0.0.1-PREVIEW</version>
 	<packaging>maven-plugin</packaging>
 	<name>azure-relay</name>
 	<description>Java library for Azure Relay</description>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>com.microsoft.azure</groupId>
 	<artifactId>azure-relay</artifactId>
-	<version>0.0.1-PREVIEW</version>
+	<version>0.0.2-PREVIEW</version>
 	<packaging>maven-plugin</packaging>
 	<name>azure-relay</name>
 	<description>Java library for Azure Relay</description>
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -15,7 +15,6 @@ import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.websocket.api.UpgradeException;
 
 class ClientWebSocket extends Endpoint implements RelayTraceSource {
-	private final AsyncLock socketLock;
 	private final AutoShutdownScheduledExecutor executor;
 	private final WebSocketContainer container = ContainerProvider.getWebSocketContainer();
 	private final TrackingContext trackingContext;
@@ -36,7 +35,6 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 		this.fragmentQueue = new InputQueue<MessageFragment>(this.executor);
 		this.closeReason = null;
 		this.trackingContext = trackingContext;
-		this.socketLock = new AsyncLock(HybridConnectionListener.EXECUTOR);
 	}
 	
 	public TrackingContext getTrackingContext() {
@@ -225,36 +223,37 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 				RelayLogger.logEvent("writingBytes", this, mode.toString());
 				
 				// The websocket API will throw if multiple sends are attempted on the same websocket simultaneously
-				// Therefore a lock is used on the send operation
-				return this.socketLock.acquireThenCompose(timeout, () -> {
-					return CompletableFutureUtil.timedRunAsync(timeout, () -> {
-						try {
-							if (mode.equals(WriteMode.TEXT)) {
-								String text = data.toString();
-								remote.sendText(text, isEnd);
-								RelayLogger.logEvent("doneWritingBytes", this, String.valueOf(text.length()));
+				return CompletableFutureUtil.timedRunAsync(timeout, () -> {
+					try {
+						if (mode.equals(WriteMode.TEXT)) {
+							String text = data.toString();
+							remote.sendText(text, isEnd);
+							RelayLogger.logEvent("writingBytesFinished", this, String.valueOf(text.length()));
+						}
+						else {
+							ByteBuffer bytes = null;
+							if (data instanceof byte[]) {
+								bytes = ByteBuffer.wrap((byte[]) data);
+							} 
+							else if (data instanceof ByteBuffer) {
+								bytes = (ByteBuffer) data;
+							}
+							else if (data instanceof String) {
+								bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
 							}
 							else {
-								ByteBuffer bytes = null;
-								if (data instanceof byte[]) {
-									bytes = ByteBuffer.wrap((byte[]) data);
-								} 
-								else if (data instanceof ByteBuffer) {
-									bytes = (ByteBuffer) data;
-								}
-								else if (data instanceof String){
-									bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
-								}
-								
-								int bytesToSend = bytes.remaining();
-								remote.sendBinary(bytes, isEnd);
-								RelayLogger.logEvent("doneWritingBytes", this, String.valueOf(bytesToSend));
+								throw new IllegalArgumentException(
+										"The data to be sent should be String, ByteBuffer or byte[], but received " + data.getClass().getSimpleName());
 							}
-						} catch (Exception e) {
-							throw RelayLogger.throwingException(e, this);
+							
+							int bytesToSend = bytes.remaining();
+							remote.sendBinary(bytes, isEnd);
+							RelayLogger.logEvent("writingBytesFinished", this, String.valueOf(bytesToSend));
 						}
-					}, executor);
-				});
+					} catch (Exception e) {
+						throw RelayLogger.throwingException(e, this);
+					}
+				}, executor);
 			}
 		}
 		else {
@@ -285,18 +284,16 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 			return this.closeTask;
 		}
 
-		this.socketLock.acquireThenApply(null, () -> {
-			try {
-				if (reason != null) {
-					this.session.close(reason);
-				} else {
-					this.session.close();
-				}
-			} catch (Throwable e) {
-				this.closeTask.completeExceptionally(e);
+		try {
+			if (reason != null) {
+				this.session.close(reason);
+			} else {
+				this.session.close();
 			}
-			return null;
-		});
+		} catch (Throwable e) {
+			this.closeTask.completeExceptionally(e);
+		}
+
 		return this.closeTask;
 	}
 	
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -7,7 +7,6 @@ import java.util.LinkedList;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.io.IOException;
-import java.io.Writer;
 
 import javax.websocket.*;
 
@@ -16,6 +15,7 @@ import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.websocket.api.UpgradeException;
 
 class ClientWebSocket extends Endpoint implements RelayTraceSource {
+	private final AsyncLock socketLock;
 	private final AutoShutdownScheduledExecutor executor;
 	private final WebSocketContainer container = ContainerProvider.getWebSocketContainer();
 	private final TrackingContext trackingContext;
@@ -26,7 +26,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	private InputQueue<String> textQueue;
 	private CompletableFuture<Void> closeTask;
 	private String cachedString;
-	
+
 	/**
 	 * Creates a websocket instance
 	 */
@@ -36,6 +36,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 		this.fragmentQueue = new InputQueue<MessageFragment>(this.executor);
 		this.closeReason = null;
 		this.trackingContext = trackingContext;
+		this.socketLock = new AsyncLock(HybridConnectionListener.EXECUTOR);
 	}
 	
 	public TrackingContext getTrackingContext() {
@@ -200,7 +201,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	 * @throws TimeoutException Throws when the sending task does not complete within the given timeout.
 	 */
 	public CompletableFuture<Void> writeAsync(Object data, Duration timeout) {
-		return writeAsync(data, timeout, WriteMode.BINARY);
+		return writeAsync(data, timeout, true, WriteMode.BINARY);
 	}
 	
 	/**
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
 
 	<groupId>com.microsoft.azure</groupId>
 	<artifactId>azure-relay</artifactId>
-	<version>0.0.1-PREVIEW</version>
+	<version>0.0.2-PREVIEW</version>
```
dlstucki: >.2 [](start = 13, length = 2)

Are you just about to create a new release? That's typically when we increase the version--as a last step before publishing.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -36,6 +36,7 @@ public ClientWebSocket(TrackingContext trackingContext, AutoShutdownScheduledExe
 		this.fragmentQueue = new InputQueue<MessageFragment>(this.executor);
 		this.closeReason = null;
 		this.trackingContext = trackingContext;
+		this.socketLock = new AsyncLock(HybridConnectionListener.EXECUTOR);
```
dlstucki: >HybridConnectionListener [](start = 34, length = 24)

I thought we ran into deadlock issues when the send side used the same thread pool as the receiver side. ClientWebSocket is used by both, correct?

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -200,60 +201,60 @@ boolean isOpen() {
...
-							RelayLogger.logEvent("writingBytes", this, "binary");
-							if (data instanceof byte[]) {
-								bytes = ByteBuffer.wrap((byte[]) data);
-							} 
-							else if (data instanceof ByteBuffer) {
-								bytes = (ByteBuffer) data;
+				RemoteEndpoint.Basic remote = this.session.getBasicRemote();
+				RelayLogger.logEvent("writingBytes", this, mode.toString());
+				
+				// The websocket API will throw if multiple sends are attempted on the same websocket simultaneously
```
dlstucki: >The websocket API will throw if multiple sends are attempted on the same websocket simultaneously [](start = 7, length = 97)

I'm curious if it's the libraries job to ensure users don't do this bad thing or the user's responsibility. My inclination is that if we try to do it for them then they don't get any guarantees as to the order of their data. If we don't do it for them, then they must wait until the previous write finishes before making the next write and their data is actually guaranteed to be in the right order.

Basically this doesn't guarantee ordering:
webSocket.writeAsync( ...) ; // bytes 0-9, NO WAIT
webSocket.writeAsync(...); // bytes 10-19, NO WAIT

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -200,60 +201,60 @@ boolean isOpen() {
...
+								else if (data instanceof ByteBuffer) {
+									bytes = (ByteBuffer) data;
+								}
+								else if (data instanceof String){
+									bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
+								}
+								
+								int bytesToSend = bytes.remaining();
+								remote.sendBinary(bytes, isEnd);
+								RelayLogger.logEvent("doneWritingBytes", this, String.valueOf(bytesToSend));
```
dlstucki: >String.valueOf(bytesToSend) [](start = 55, length = 27)

Does this log the actual customer data? We shouldn't be doing that without them turning it on _very_ explicitly.

src/main/java/com/microsoft/azure/relay/RelayLogger.java
```diff
@@ -128,6 +128,7 @@ private static void init() {
 		map.put("httpSendResponse", "%s: Sending the response command on the %s connection, status: %s.");
 		map.put("httpSendResponseFinished", "%s: Finished sending the response command on the %s connection, status: %s.");
 		map.put("httpUserRequestHandlerException", "%s: exception in the user's request handler.");
```
dlstucki: >exception in the user's request handler. [](start = 50, length = 40)

I know you're not touching this line but I happened to look at it anyway...

In the case of user request handler error do no details of the exception get logged?

If something more is needed feel free to file an issue and resolve this comment.

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -209,24 +209,27 @@ void invokeRequestHandler(RequestCommandAndStream requestAndStream) {
...
 			ListenerCommand listenerCommand = new ListenerCommand(null);
 			listenerCommand.setResponse(responseCommand);
 			String command = listenerCommand.getResponse().toJsonString();
 
 			// We need to respond over the rendezvous connection
-			CompletableFuture<Void> sendCommandTask = this.rendezvousWebSocket.writeAsync(command, timeout, WriteMode.TEXT)
-					.thenAccept(bytesWritten -> {
-						RelayLogger.logEvent("httpSendResponseFinished", this, "rendezvous", String.valueOf(responseCommand.getStatusCode()));
-					});
+			CompletableFuture<Void> sendCommandTask = this.ensureRendezvousAsync(timeRemaining.remainingTime())
```
dlstucki: >ensureRendezvousAsync [](start = 50, length = 21)

I wonder if the fact that this else block is only hit when `this.rendezvousWebSocket != null` means ensureRendezvousAsync can be skipped?

dlstucki: It appears my statement is not true. Since ensureRendezvousAsync assigns to this.rendezvousWebSocket before it is connected, and not after what I had hoped to be the case won't work.

---
In reply to: [291718830](https://github.com/Azure/azure-relay-java/pull/45#discussion_r291718830) [](ancestors = 291718830)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -293,4 +311,19 @@ private static void httpRequestSender(String method, String msgExpected, String
...
 		assertEquals("Http connection sender did not receive the expected response message.", msgExpected, builder.toString());
 	}
+	
+	private static void httpRequestHandlerMultipleSend(RelayedHttpListenerContext context, String firstMsg, String secondMsg) {
+		RelayedHttpListenerResponse response = context.getResponse();
+		response.setStatusCode(STATUS_CODE);
+		response.setStatusDescription(STATUS_DESCRIPTION);
+		
+		try {
+			response.getOutputStream().write(firstMsg.getBytes());
```
dlstucki: >write [](start = 30, length = 5)

Does write call writeAsync and wait for it to complete or just kick it off?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -204,6 +204,24 @@ public void httpLargePOSTAndLargeResponseTest() throws IOException {
         httpRequestSender("POST", largeStr, largeStr);
 	}
 	
+	@Test
+	public void httpWriteSmallThenSmallResponseTest() throws IOException {
```
dlstucki: >httpWriteSmallThenSmallResponseTest [](start = 13, length = 35)

Does anything in the test suite fail without the send async lock that's being introduced?

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
dlstucki: Or is it to solve the issue where user calls writeAsync the closeAsync without waiting?

CF writeFuture = ws.writeAsync(...);
CF closeFuture = ws.closeAsync(); // close waits for inflight send to drain?

---
In reply to: [291712180](https://github.com/Azure/azure-relay-java/pull/45#discussion_r291712180) [](ancestors = 291712180)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -200,60 +201,60 @@ boolean isOpen() {
...
-							} 
-							else if (data instanceof ByteBuffer) {
-								bytes = (ByteBuffer) data;
+				RemoteEndpoint.Basic remote = this.session.getBasicRemote();
+				RelayLogger.logEvent("writingBytes", this, mode.toString());
+				
+				// The websocket API will throw if multiple sends are attempted on the same websocket simultaneously
+				// Therefore a lock is used on the send operation
+				return this.socketLock.acquireThenCompose(timeout, () -> {
+					return CompletableFutureUtil.timedRunAsync(timeout, () -> {
```
dlstucki: >timeout [](start = 48, length = 7)

I suppose in a case like this a TimeoutHelper should be used. Time waiting for the socketLock should be deducted from the remaining operation time.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -200,60 +201,60 @@ boolean isOpen() {
...
+								ByteBuffer bytes = null;
+								if (data instanceof byte[]) {
+									bytes = ByteBuffer.wrap((byte[]) data);
+								} 
+								else if (data instanceof ByteBuffer) {
+									bytes = (ByteBuffer) data;
+								}
+								else if (data instanceof String){
+									bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
+								}
```
dlstucki: Isn't it possible for an Object to not fall into any of these if/else-if combinations? Ideally this API should throw instead of doing an empty send in that case.

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -324,13 +327,20 @@ public void setWriteTimeout(Duration writeTimeout) {
...
-							.sendBytesOverRendezvousAsync(this.writeBufferStream, timeout).thenRun(() -> {
-								this.writeBufferStream.clear();
-								if (this.writeBufferFlushTimer != null) {
-									this.writeBufferFlushTimer.cancel();
-								}
-							}));
+					return sendResponseTask.thenCompose($void -> {
+						// Get a new buffer backed by the non empty segment of the write buffer array
+						this.writeBufferStream.flip();
+						ByteBuffer trimmedBuffer = ByteBuffer.wrap(
```
dlstucki: >ByteBuffer.wrap( [](start = 33, length = 16)

What benefit does wrapping the ByteBuffer add here?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: >httpWriteSmallThenSmallResponseTest [](start = 13, length = 35)

Would anything in the tests catch and fail on a scenario where the message arrived with extra '\0' characters? (Converting from bytes to String might throw away trailing '\0' chars...

"FirstSecond\0\0\0\0"

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -324,13 +327,20 @@ public void setWriteTimeout(Duration writeTimeout) {
...
+								this.writeBufferStream.array(), 
+								this.writeBufferStream.position(), 
+								this.writeBufferStream.remaining());
```
dlstucki: Can you help me understand this? Why is a clone of the writeBufferStream needed (the wrap which I believe creates something basically equivalent)?

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -324,13 +327,20 @@ public void setWriteTimeout(Duration writeTimeout) {
...
+						return this.connection.sendBytesOverRendezvousAsync(trimmedBuffer, timeout);
```
dlstucki: >timeout [](start = 73, length = 7)

Should this be using TimeoutHelper? Is it allowed to take 2*timeout?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: >httpWriteSmallThenSmallResponseTest [](start = 13, length = 35)

Are there any tests which write some bytes to the response, pausing allowing the flush timer to send those bytes, then send some more bytes in the response body and validate the exact number of bytes and values are transmitted to the client?

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -209,24 +209,27 @@ void invokeRequestHandler(RequestCommandAndStream requestAndStream) {
 			return this.listener.sendControlCommandAndStreamAsync(listenerCommand, responseBodyBuffer, timeout)
 					.thenRun(() -> RelayLogger.logEvent("httpSendResponseFinished", this, "control", String.valueOf(responseCommand.getStatusCode())));
 		} else {
+			TimeoutHelper timeRemaining = new TimeoutHelper(timeout);
```
dlstucki: >TimeoutHelper timeRemaining = new TimeoutHelper(timeout); [](start = 3, length = 57)

I wouldn't bother changing this now, but keep in mind "timeoutHelper" is really a replacement for "timeout" argument. Any code path which would use timeout should really be aware of timeoutHelper. My personal guideline is therefore to try to keep TimeoutHelper at the top of the method.

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
bainian12345: It's to wrap only the part of the buffer that has been written to, so the empty bytes in the buffer aren't sent.

---
In reply to: [292073271](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292073271) [](ancestors = 292073271)

dlstucki: Does that actually happen without the wrap here? 

---
In reply to: [292098393](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292098393) [](ancestors = 292098393,292073271)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
bainian12345: This lock is to guard the websocket from multiple writes concurrently, not related to close.

All the unit tests are doing synchronized writes as of now and not running into this issue. I will do more investigation on this later in a separate PR. Probably going to leave this lock out for now.

---
In reply to: [292064317](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292064317) [](ancestors = 292064317,291712180)

dlstucki: If a user calls writeAsync twice concurrently I think that ideally we should give them an exception. Even if no exception it's pretty much into undefined territory.

**However**, if our timer based flush is racing with the user calling another writeAsync we almost certainly need to serialize things there.

---
In reply to: [292622840](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292622840) [](ancestors = 292622840,292064317,291712180)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
bainian12345: You are right, what I needed was the flip() on the buffer, not wrapping it.

---
In reply to: [292098796](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292098796) [](ancestors = 292098796,292098393,292073271)

src/main/java/com/microsoft/azure/relay/RelayLogger.java
bainian12345: Should we log any details about user implementation? I was concerned about exposing user internal details

---
In reply to: [291715446](https://github.com/Azure/azure-relay-java/pull/45#discussion_r291715446) [](ancestors = 291715446)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: Runs fine for now with synchronized writes, will do async writes in separate PR

---
In reply to: [292063905](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292063905) [](ancestors = 292063905)

src/main/java/com/microsoft/azure/relay/RelayLogger.java
dlstucki: It's far less common to have sensitive information in Exceptions themselves. Personally, I'd to the exception type, message, callstack in the tracing. Also, these traces go wherever our customers configure them to go after they opt-in by turning on tracing. Typically that's something that stays on the same machine as the user code. One has to be more careful when sending details over the wire or saving logs in multitenant systems or in the cloud.

---
In reply to: [292631454](https://github.com/Azure/azure-relay-java/pull/45#discussion_r292631454) [](ancestors = 292631454,291715446)

PR #46, opened 2019-06-12
Allow writeAsync() for users in HybridHttpConnection.ResponseStream

bainian12345:
- Changed the type of output stream from HTTP context object to expose writeAsync()
- Slightly modified overrides of write methods from ResponseStream to resemble OutputStream signatures
- Refactored tests

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
@@ -1,6 +1,7 @@
 package com.microsoft.azure.relay;
 
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -279,7 +280,7 @@ class HybridHttpConnection implements RelayTraceSource {
 	}
 
 	final class ResponseStream extends OutputStream {
-		private static final long WRITE_BUFFER_FLUSH_TIMEOUT_MILLIS = 2000;
+		static final long WRITE_BUFFER_FLUSH_TIMEOUT_MILLIS = 2000;
 		private final HybridHttpConnection connection;
 		private final RelayedHttpListenerContext context;
 		private final AsyncLock asyncLock;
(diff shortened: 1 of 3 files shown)


src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -304,38 +296,50 @@ private static void httpRequestSender(String method, String msgExpected, String
...
 		response.setStatusCode(STATUS_CODE);
 		response.setStatusDescription(STATUS_DESCRIPTION);
 		
+		CompletableFuture<?>[] cfs = new CompletableFuture<?>[messages.length];
 		try {
-			response.getOutputStream().write(firstMsg.getBytes());
-			if (hasPause) {
-				Thread.sleep(2500);
+			for (int i = 0; i < messages.length; i++) {
+				cfs[i] = response.getOutputStream().writeAsync(messages[i].getBytes(), 0, messages[i].length());
```
dlstucki: >writeAsync [](start = 40, length = 10)

So this calls writeAsync 'i' times but doesn't wait for the previous writeAsync to finish? IMHO that's asking for undefined behavior and customers should't be doing this.

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -304,38 +296,50 @@ private static void httpRequestSender(String method, String msgExpected, String
...
+	
+	private static String readStringFromStream(InputStream inputStream) throws IOException {
+		if (inputStream == null) {
+			return emptyStr;
+		}
+		
+		StringBuilder builder = new StringBuilder();
+		String inputLine;
+		try (BufferedReader inStream = new BufferedReader(new InputStreamReader(inputStream))){
+			while ((inputLine = inStream.readLine()) != null) {
```
dlstucki: >readLine [](start = 32, length = 8)

I feel like I've asked this before. What happens with multi-line strings? Does '\n' get removed from the inputStream in this process? What about '\r\n' in the original data? I hope we have some tests which are using boring-old bytes to make sure the SDK isn't messing with the user data in any way at all.

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -346,21 +347,58 @@ public void setWriteTimeout(Duration writeTimeout) {
...
+		/**
+		 * Writes b.length bytes from the specified byte array to this response stream.
+		 */
+		@Override
+		public void write(byte[] b) throws IOException {
+			try {
+				this.writeAsync(b, 0, b.length).join();
+			} catch (CompletionException e) {
+				throw new IOException(e.getCause());
+			}
```
dlstucki: Can this simply be a one line call to the other **sync** overload?

this.write(b, 0, b.length);

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: Because we have locks on writeAsync, in reality they won't run into the simultaneous write exception thrown from ClientWebSocket. Do you think we should explicitly throw when user writes without waiting then?

---
In reply to: [293086335](https://github.com/Azure/azure-relay-java/pull/46#discussion_r293086335) [](ancestors = 293086335)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: I will check on that in a separate PR

---
In reply to: [293099146](https://github.com/Azure/azure-relay-java/pull/46#discussion_r293099146) [](ancestors = 293099146)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
bainian12345: Good idea

---
In reply to: [293102083](https://github.com/Azure/azure-relay-java/pull/46#discussion_r293102083) [](ancestors = 293102083)

src/main/java/com/microsoft/azure/relay/HybridHttpConnection.java
```diff
@@ -346,21 +354,46 @@ public void setWriteTimeout(Duration writeTimeout) {
...
 		}
 
-		public void write(String text) {
-			this.writeAsync(text.getBytes(StringUtil.UTF8), 0, text.length()).join();
+		/**
+		 * Writes the given text to this response stream.
+		 */
+		public void write(String text) throws IOException {
+			this.write(text.getBytes(StringUtil.UTF8), 0, text.length());
```
dlstucki: >text.length() [](start = 49, length = 13)

Sorry I didn't catch this before.

text.length() gives the number of characters which may not be the same length as the number of bytes. If this overload is truly desired (I wonder if we could do without it) then it needs to capture the byte[] from text.getBytes() and use the length of that array. Also, I suppose it should be documented that this serializes a String using UTF-8 encoding.

dlstucki: This may have existed before the PR any how and could be addressed in a future PR.

---
In reply to: [293595994](https://github.com/Azure/azure-relay-java/pull/46#discussion_r293595994) [](ancestors = 293595994)

bainian12345: will remove this overload

---
In reply to: [293596279](https://github.com/Azure/azure-relay-java/pull/46#discussion_r293596279) [](ancestors = 293596279,293595994)

PR #47, opened 2019-06-14
Refactored and added tests to verify sending and receiving of all byte values

bainian12345:
Refactored and added tests to verify sending and receiving of all byte values

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -247,6 +247,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 							}
 							
 							int bytesToSend = bytes.remaining();
+							// sendBinary() will cause the content of the byte array within the ByteBuffer to change
 							remote.sendBinary(bytes, isEnd);
 							RelayLogger.logEvent("writingBytesFinished", this, String.valueOf(bytesToSend));
 						}
--- a/src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
+++ b/src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
@@ -31,33 +31,25 @@ import org.junit.Test;
 import com.microsoft.azure.relay.HybridHttpConnection.ResponseStream;
 
 public class SendReceiveTest {
+	private static final int EMPTY_BYTES_SIZE = 0;
+	private static final int SMALL_BYTES_SIZE = 256;
+	private static final int LARGE_BYTES_SIZE = 64*1024 + 1;
+	private static final int STATUS_CODE = HttpStatus.ACCEPTED_202;
+	private static final String STATUS_DESCRIPTION = "OK";
+	private static final byte[] EMPTY_BYTES = getTestBytes(EMPTY_BYTES_SIZE);
+	private static final byte[] SMALL_BYTES = getTestBytes(SMALL_BYTES_SIZE);
+	private static final byte[] LARGE_BYTES = getTestBytes(LARGE_BYTES_SIZE);
+	
 	private static HybridConnectionListener listener;
 	private static TokenProvider tokenProvider;
 	private static HybridConnectionClient client;
-	private static final int STATUS_CODE = HttpStatus.ACCEPTED_202;
-	private static final String STATUS_DESCRIPTION = "OK";
 	private static int listenerReceiveCount = 0;
 	private static int senderReceiveCount = 0;
-	
-	// empty test string for GET method
-	private static String emptyStr = "";
-	// small test string that's under 64kb
-	private static String smallStr = "smallStr";
-	// large test string that's over 64kb
-	private static String largeStr;
 	
 	@BeforeClass
 	public static void init() throws URISyntaxException {
 		tokenProvider = TokenProvider.createSharedAccessSignatureTokenProvider(TestUtil.KEY_NAME, TestUtil.KEY);
 		listener = new HybridConnectionListener(new URI(TestUtil.RELAY_NAMESPACE_URI + TestUtil.ENTITY_PATH), tokenProvider);
-
-		// Build the large string in small chunks because hardcoding a large string may not compile
-		StringBuilder builder = new StringBuilder();
-		String alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
-		for (int i = 0; i < 2000; i++) {
-			builder.append(alphabet);
-		}
-		largeStr = builder.toString();
 		listener.openAsync(Duration.ofSeconds(15)).join();
 	}
 	
@@ -75,32 +67,32 @@ public class SendReceiveTest {
 	
 	@Test
 	public void websocketSmallSendSmallResponseTest() {
-		CompletableFuture<Void> listenerTask = websocketListener(smallStr, smallStr);
-		CompletableFuture<Void> clientTask = websocketClient(smallStr, smallStr);
+		CompletableFuture<Void> listenerTask = sendAndReceiveWithWebsocketListener(SMALL_BYTES, SMALL_BYTES);
+		CompletableFuture<Void> clientTask = sendAndReceiveWithWebsocketClient(SMALL_BYTES, SMALL_BYTES);
 		listenerTask.join();
 		clientTask.join();
 	}
 	
 	@Test
 	public void websocketSmallSendLargeResponseTest() {
-		CompletableFuture<Void> listenerTask = websocketListener(smallStr, largeStr);
-		CompletableFuture<Void> clientTask = websocketClient(largeStr, smallStr);
+		CompletableFuture<Void> listenerTask = sendAndReceiveWithWebsocketListener(SMALL_BYTES, LARGE_BYTES);
+		CompletableFuture<Void> clientTask = sendAndReceiveWithWebsocketClient(LARGE_BYTES, SMALL_BYTES);
 		listenerTask.join();
 		clientTask.join();
 	}
 	
 	@Test
 	public void websocketLargeSendSmallResponseTest() {
-		CompletableFuture<Void> listenerTask = websocketListener(largeStr, smallStr);
-		CompletableFuture<Void> clientTask = websocketClient(smallStr, largeStr);
+		CompletableFuture<Void> listenerTask = sendAndReceiveWithWebsocketListener(LARGE_BYTES, SMALL_BYTES);
+		CompletableFuture<Void> clientTask = sendAndReceiveWithWebsocketClient(SMALL_BYTES, LARGE_BYTES);
 		listenerTask.join();
 		clientTask.join();
 	}
 	
 	@Test
 	public void websocketLargeSendLargeResponseTest() {
-		CompletableFuture<Void> listenerTask = websocketListener(largeStr, largeStr);
-		CompletableFuture<Void> clientTask = websocketClient(largeStr, largeStr);
+		CompletableFuture<Void> listenerTask = sendAndReceiveWithWebsocketListener(LARGE_BYTES, LARGE_BYTES);
+		CompletableFuture<Void> clientTask = sendAndReceiveWithWebsocketClient(LARGE_BYTES, LARGE_BYTES);
 		listenerTask.join();
 		clientTask.join();
 	}
@@ -111,19 +103,19 @@ public class SendReceiveTest {
 		CompletableFuture<Integer> senderReceiveCount = new CompletableFuture<Integer>();
 		CompletableFuture<Integer> listenerReceiveCount = new CompletableFuture<Integer>();
 		
-		listener.acceptConnectionAsync().thenAcceptAsync((websocket) -> {
+		listener.acceptConnectionAsync().thenAcceptAsync(websocket -> {
 			for (int i = 1; i <= timesToRepeat; i++) {
 				websocket.readAsync().join();
-				websocket.writeAsync(StringUtil.toBuffer("hi")).join();
+				websocket.writeAsync(ByteBuffer.wrap(SMALL_BYTES)).join();
 				if (i == timesToRepeat) {
 					listenerReceiveCount.complete(i);
 				}
 			}
 		});
 		
-		client.createConnectionAsync().thenAccept((clientWebSocket) -> {
+		client.createConnectionAsync().thenAccept(clientWebSocket -> {
 			for (int i = 1; i <= timesToRepeat; i++) {
-				clientWebSocket.writeAsync(StringUtil.toBuffer("hi")).join();
+				clientWebSocket.writeAsync(ByteBuffer.wrap(SMALL_BYTES)).join();
 				clientWebSocket.readAsync().join();
 				if (i == timesToRepeat) {
 					senderReceiveCount.complete(i);
@@ -146,17 +138,17 @@ public class SendReceiveTest {
 		CompletableFuture<Void> senderReceiveTask = new CompletableFuture<Void>();
 		
 		for (int i = 1; i <= numberOfSenders; i++) {
-			listener.acceptConnectionAsync().thenAccept((websocket) -> {
+			listener.acceptConnectionAsync().thenAccept(websocket -> {
 				websocket.readAsync().thenRun(() -> {
 					listenerReceiveCount++;
-					websocket.writeAsync(StringUtil.toBuffer("hi")).join();
+					websocket.writeAsync(ByteBuffer.wrap(SMALL_BYTES)).join();
 					websocket.closeAsync(new CloseReason(CloseCodes.NORMAL_CLOSURE, "Normal closure from listener")).join();
 				});
 			});
 			
 			HybridConnectionClient newClient = new HybridConnectionClient(new URI(TestUtil.RELAY_NAMESPACE_URI + TestUtil.ENTITY_PATH), tokenProvider);
-			newClient.createConnectionAsync().thenAccept((socket) -> {
-				socket.writeAsync(StringUtil.toBuffer("hi")).join();
+			newClient.createConnectionAsync().thenAccept(socket -> {
+				socket.writeAsync(ByteBuffer.wrap(SMALL_BYTES)).join();
 				socket.readAsync().thenRun(() -> {
 					senderReceiveCount++;
 					if (senderReceiveCount == numberOfSenders) {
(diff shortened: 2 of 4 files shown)


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -247,6 +247,7 @@ else if (data instanceof String) {
 							}
 							
 							int bytesToSend = bytes.remaining();
+							// sendBinary() will cause the content of the byte array within the ByteBuffer to change
```
dlstucki: >content of the byte array [](start = 38, length = 25)

This is weird. What do the array contents become?

bainian12345: Becomes garbage values...I haven't looked into it, not sure what they are doing with the array after sending

---
In reply to: [294013165](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294013165) [](ancestors = 294013165)

dlstucki: no reason, I'm just surprised they party on someone else's buffer

---
In reply to: [294013165](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294013165) [](ancestors = 294013165)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -326,18 +392,28 @@ private static void sendResponseMessages(RelayedHttpListenerContext context, Str
...
+				break;
+			} else {
+				byteStream.write(i);
 			}
 		}
-		return builder.toString();
+		return byteStream.toByteArray();
+	}
+	
+	private byte[] getByteSet() {
```
dlstucki: >getByteSet [](start = 16, length = 10)

Do all these byte-level tests use exactly 256bytes?

bainian12345: All 256 bytes are checked for equality

---
In reply to: [294017638](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294017638) [](ancestors = 294017638)

dlstucki: At some point it would be nice to check that an odd number of bytes work. 1, 255, 1024, 4k, 64k-1, 64k, 64k+1, etc. In the future consider having getByteSet take a size and fill an array of that size with this repeating byte pattern.

---
In reply to: [294017638](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294017638) [](ancestors = 294017638)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -75,29 +76,33 @@ public void createClient() throws URISyntaxException {
...
+		CompletableFuture<Void> clientTask = websocketClient(largeStr, smallStr);
 		listenerTask.join();
+		clientTask.join();
 	}
 	
 	@Test
 	public void websocketLargeSendSmallResponseTest() {
 		CompletableFuture<Void> listenerTask = websocketListener(largeStr, smallStr);
-		websocketClient(smallStr, largeStr);
+		CompletableFuture<Void> clientTask = websocketClient(smallStr, largeStr);
```
dlstucki: >websocketClient [](start = 39, length = 15)

nothing to do in this PR, I just want to point out this method name is a noun only and method names are more self-documenting when they contain verbs too.

dlstucki: e.g. RunClientTest, TestClient, ProcessClient, ConfigureClient, etc.

---
In reply to: [294378342](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294378342) [](ancestors = 294378342)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -170,6 +175,34 @@ public void websocketMultipleSenderTest() throws URISyntaxException {
...
+					byte[] byteSet = getByteSet();
+					assertTrue("Websocket listener did not receive the expected bytes.", Arrays.equals(byteSet, bufferReceived.array()));
+					return connection.writeAsync(ByteBuffer.wrap(byteSet));
+				})
+				.thenCompose($void -> {
+					return connection.closeAsync(new CloseReason(CloseCodes.NORMAL_CLOSURE, "Normal closure from listener"));
+				});
+		});
+		
+		HybridConnectionClient newClient = new HybridConnectionClient(new URI(TestUtil.RELAY_NAMESPACE_URI + TestUtil.ENTITY_PATH), tokenProvider);
```
dlstucki: >newClient [](start = 25, length = 9)

should this test ideally close the HCClient no matter what?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -240,6 +273,32 @@ public void httpWriteResponseEnsureOrderTest() throws IOException {
 		httpRequestSender("POST", String.join("", messages), smallStr);
 	}
 	
+	@Test
+	public void httpWriteReceiveAllByteValuesTest() throws IOException {
+		byte[] byteSet = getByteSet();
+		listener.setRequestHandler(context -> {
+			RelayedHttpListenerResponse response = context.getResponse();
+			response.setStatusCode(STATUS_CODE);
```
dlstucki: >setStatusCode [](start = 12, length = 13)

Are there any tests in this suite which use any other status code?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -240,6 +273,32 @@ public void httpWriteResponseEnsureOrderTest() throws IOException {
...
+			response.setStatusDescription(STATUS_DESCRIPTION);
+			
+			try {
+				response.getOutputStream().writeAsync(byteSet, 0, byteSet.length).join();
+			} catch (Exception e) {
+				fail(e.getMessage());
```
dlstucki: >fail [](start = 4, length = 4)

If this fail occurs in the requestHandler callback does the unit test fail? (Actually, I think so because the correct response most likely won't appear.

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -240,6 +273,32 @@ public void httpWriteResponseEnsureOrderTest() throws IOException {
...
+			} finally {
+			    context.getResponse().close();
```
dlstucki: >context.getResponse() [](start = 7, length = 21)

could this be "response.close()"?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -280,29 +339,36 @@ private static void httpRequestHandler(RelayedHttpListenerContext context, Strin
 	}
 	
 	private static void httpRequestSender(String method, String msgExpected, String msgToSend) throws IOException {
+		HttpURLConnection connection = getHttpConnection();
+		sendBytesOverHttp(connection, msgToSend.getBytes(), 0, msgToSend.length());
```
dlstucki: >msgToSend.length() [](start = 57, length = 18)

call msgToSend.getBytes() (UTF-8?), hang on to the byte[] and use its length here instead of String.length() (the UTF-16 character count).

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -280,29 +339,36 @@ private static void httpRequestHandler(RelayedHttpListenerContext context, Strin
...
-		conn.setDoOutput(true);
-		OutputStreamWriter out = new OutputStreamWriter(conn.getOutputStream());
-		out.write(message, 0, message.length());
+		HttpURLConnection connection = (HttpURLConnection)url.openConnection();
+		connection.setRequestProperty("ServiceBusAuthorization", tokenString);
+		return connection;
+	}
+	
+	private static void sendBytesOverHttp(HttpURLConnection connection, byte[] b, int off, int len) throws IOException {
+		connection.setRequestMethod((len > 0) ? "POST" : "GET");
```
dlstucki: >"GET" [](start = 51, length = 5)

For future consideration: Do the following verbs work in our Java Library?

Delete, Get, Head, Options, Post, Put, Trace, Patch

dlstucki: >( [](start = 30, length = 1)

total style nit but consider omitting these parenthesis if they're not needed.

dlstucki: I said this then I started to feel a bit silly. (At this point you have more Java experience on the team than I do and perhaps the Java conventions should come from you). Feel free to ignore. We do have a .NET StyleCop rule which enforces this in C# though. 

---
In reply to: [294385721](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294385721) [](ancestors = 294385721)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -326,18 +392,28 @@ private static void sendResponseMessages(RelayedHttpListenerContext context, Str
 		}
 	}
 	
-	private static String readStringFromStream(InputStream inputStream) throws IOException {
+	private static byte[] readBytesFromStream(InputStream inputStream) throws IOException {
 		if (inputStream == null) {
-			return emptyStr;
+			return new byte[0];
```
dlstucki: > new byte[0]; [](start = 9, length = 13)

Food for thought: I'm curious if this has the result of converting a null inputStream, kind of like (String)null, into a "". Does this make it harder to test any variations we might be interested in?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: Do you mean replace the current tests using small/large strings with these bytes repeated? I'm not sure what kind of difference it would make with different sizes like 1024 vs 4096

---
In reply to: [294377702](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294377702) [](ancestors = 294377702,294017638)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: HCClient has no close(), closing the connection it creates instead.

---
In reply to: [294380700](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294380700) [](ancestors = 294380700)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: Do you feel like there should be separate tests that just focus on the status code and description? I didn't include any because I didn't think there are any special values that we need to test specifically

---
In reply to: [294381396](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294381396) [](ancestors = 294381396)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: First of all I'm trying to get you outside of testing strings. The library deals with bytes not Strings as near AFAICT. Strings convert to bytes in a MANY different ways (which are outside the scope of the library here).

I want to ensure if the library is given these 11 bytes:
0xFF 0xFE 0x65 0x00 0x66 0x00 0x67 0x00 0x00 0x00 0x00 (UTF-16LE with byte order mark string of "ABC" plus extra 0x00 0x00 terminator plus one extra 0x00 byte to make it an odd number)

that our library doesn't convert it to something else like these 4 bytes which once converted to String are equivalent to the original 10 bytes:
0x65 0x66 0x67 0x00            (ASCII or UTF-8 with no byte order mark, lost trailing 0x00 bytes).

---
In reply to: [295527790](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295527790) [](ancestors = 295527790,294377702,294017638)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: I'd feel better the tests do at least 1 variation (2 total) on StatusCode. Otherwise how do you know it's actually using the one you told it to? Same thing with Description.

---
In reply to: [295535676](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295535676) [](ancestors = 295535676,294381396)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: I did this to avoid potentially having null checks everywhere.

The only side effect I can think of is if we send and receive null vs. byte[0], but according to the code, sending null would just return and do nothing, so I don't think we need to worry about tests sending and receiving null vs. byte[0].

---
In reply to: [294387162](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294387162) [](ancestors = 294387162)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: Will include in future PR

---
In reply to: [294385150](https://github.com/Azure/azure-relay-java/pull/47#discussion_r294385150) [](ancestors = 294385150)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -106,19 +103,19 @@ public void websocketRepeatedSendReceiveTest() {
 		CompletableFuture<Integer> senderReceiveCount = new CompletableFuture<Integer>();
 		CompletableFuture<Integer> listenerReceiveCount = new CompletableFuture<Integer>();
 		
-		listener.acceptConnectionAsync().thenAcceptAsync((websocket) -> {
+		listener.acceptConnectionAsync().thenAcceptAsync(websocket -> {
 			for (int i = 1; i <= timesToRepeat; i++) {
 				websocket.readAsync().join();
-				websocket.writeAsync(StringUtil.toBuffer("hi")).join();
+				websocket.writeAsync(ByteBuffer.wrap(SMALL_BYTES)).join();
```
dlstucki: >ByteBuffer.wrap(SMALL_BYTES)). [](start = 25, length = 30)

What impact does this comment from ClientWebSocket.java have on the shared SMALL_BYTES array?
"sendBinary() will cause the content of the byte array within the ByteBuffer to change"

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -172,147 +169,148 @@ public void websocketMultipleSenderTest() throws URISyntaxException {
...
 		});
 	}
 	
-	private static CompletableFuture<Void> websocketListener(String msgExpected, String msgToSend) {
-		return listener.acceptConnectionAsync().thenComposeAsync((channel) -> {
+	private static CompletableFuture<Void> sendAndReceiveWithWebsocketListener(byte[] msgExpected, byte[] msgToSend) {
+		return listener.acceptConnectionAsync().thenComposeAsync(channel -> {
 			return channel.readAsync().thenAccept(bytesReceived -> {
-				assertEquals("Websocket listener did not receive the expected message.", msgExpected, new String(bytesReceived.array()));
+				assertTrue("Websocket listener did not receive the expected message.", Arrays.equals(msgExpected, bytesReceived.array()));
```
dlstucki: >bytesReceived.array() [](start = 102, length = 21)

If the ByteBuffer ever changes to not happen to give you a perfectly sized array this will fail, right?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: is that why some places call .clone()?

---
In reply to: [295925633](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295925633) [](ancestors = 295925633)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: That makes sense, I will try to take that into account probably in a future PR.

---
In reply to: [295557975](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295557975) [](ancestors = 295557975,295535676,294381396)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: Yes, `.clone()` just to avoid the content getting messed up

---
In reply to: [295927864](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295927864) [](ancestors = 295927864,295925633)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -106,19 +103,19 @@ public void websocketRepeatedSendReceiveTest() {
...
 					listenerReceiveCount.complete(i);
 				}
 			}
 		});
 		
-		client.createConnectionAsync().thenAccept((clientWebSocket) -> {
+		client.createConnectionAsync().thenAccept(clientWebSocket -> {
 			for (int i = 1; i <= timesToRepeat; i++) {
-				clientWebSocket.writeAsync(StringUtil.toBuffer("hi")).join();
+				clientWebSocket.writeAsync(ByteBuffer.wrap(SMALL_BYTES)).join();
```
dlstucki: >SMALL_BYTES [](start = 47, length = 11)

clone here?

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: Let's put clone into our ClientWebSocket then.

---
In reply to: [295930809](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295930809) [](ancestors = 295930809,295927864,295925633)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
bainian12345: It would fail, but currently `readAsync` returns a perfectly wrapped ByteBuffer. It's not like we are letting the user bring their own buffer anyways as of now

---
In reply to: [295927555](https://github.com/Azure/azure-relay-java/pull/47#discussion_r295927555) [](ancestors = 295927555)

PR #48, opened 2019-06-20
Clone bytes in ClientWebSocket before sending to prevent content being modified

bainian12345:
Clone bytes in ClientWebSocket before sending to prevent content being modified

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -231,21 +231,23 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 							RelayLogger.logEvent("writingBytesFinished", this, String.valueOf(text.length()));
 						}
 						else {
-							ByteBuffer bytes = null;
+							byte[] bytes;
 							if (data instanceof byte[]) {
-								bytes = ByteBuffer.wrap(((byte[]) data).clone());
+								bytes = ((byte[]) data).clone();
 							} 
 							else if (data instanceof ByteBuffer) {
-								bytes = deepCopyByteBuffer((ByteBuffer) data);
+								ByteBuffer buffer = (ByteBuffer) data;
+								bytes = new byte[buffer.remaining()];
+								buffer.get(bytes);
 							}
 							else {
 								throw new IllegalArgumentException(
 									"The data to be sent should be ByteBuffer or byte[], but received " + data.getClass().getSimpleName());
 							}
 							
-							int bytesToSend = bytes.remaining();
+							int bytesToSend = bytes.length;
 							// sendBinary() will cause the content of the byte array within the ByteBuffer to change
-							remote.sendBinary(bytes, isEnd);
+							remote.sendBinary(ByteBuffer.wrap(bytes), isEnd);
 							RelayLogger.logEvent("writingBytesFinished", this, String.valueOf(bytesToSend));
 						}
 					} catch (Exception e) {
@@ -339,15 +341,6 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	@OnError
 	public void onError(Session session, Throwable cause) {
 		RelayLogger.throwingException(cause, this);
-	}
-	
-	private static ByteBuffer deepCopyByteBuffer(ByteBuffer original) {
-		ByteBuffer clone = ByteBuffer.allocate(original.capacity());
-		clone.put(original.array());
-		clone.position(original.position());
-		clone.limit(original.limit());
-		clone.order(original.order());
-    	return clone;
 	}
 	
 	private static class MessageFragment {
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -233,17 +233,14 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 						else {
 							ByteBuffer bytes = null;
 							if (data instanceof byte[]) {
-								bytes = ByteBuffer.wrap((byte[]) data);
+								bytes = ByteBuffer.wrap(((byte[]) data).clone());
 							} 
 							else if (data instanceof ByteBuffer) {
-								bytes = (ByteBuffer) data;
-							}
-							else if (data instanceof String) {
-								bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
+								bytes = deepCopyByteBuffer((ByteBuffer) data);
 							}
 							else {
 								throw new IllegalArgumentException(
-										"The data to be sent should be String, ByteBuffer or byte[], but received " + data.getClass().getSimpleName());
+									"The data to be sent should be ByteBuffer or byte[], but received " + data.getClass().getSimpleName());
 							}
 							
 							int bytesToSend = bytes.remaining();
@@ -342,6 +339,15 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	@OnError
 	public void onError(Session session, Throwable cause) {
 		RelayLogger.throwingException(cause, this);
+	}
+	
+	private static ByteBuffer deepCopyByteBuffer(ByteBuffer original) {
+		ByteBuffer clone = ByteBuffer.allocate(original.capacity());
+		clone.put(original.array());
+		clone.position(original.position());
+		clone.limit(original.limit());
+		clone.order(original.order());
+    	return clone;
 	}
 	
 	private static class MessageFragment {
--- a/src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
+++ b/src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
@@ -72,6 +72,18 @@ public class SendReceiveTest {
 		listenerTask.join();
 		clientTask.join();
 	}
+
+	@Test
+	public void websocketSmallSendSmallResponseByteBufferTest() {
+		CompletableFuture<Void> listenerTask = sendAndReceiveWithWebsocketListener(
+			ByteBuffer.wrap(SMALL_BYTES.clone(), 10, SMALL_BYTES_SIZE / 2),
+			ByteBuffer.wrap(SMALL_BYTES.clone(), 2, SMALL_BYTES_SIZE - 20));
+		CompletableFuture<Void> clientTask = sendAndReceiveWithWebsocketClient(
+			ByteBuffer.wrap(SMALL_BYTES.clone(), 2, SMALL_BYTES_SIZE - 20),
+			ByteBuffer.wrap(SMALL_BYTES.clone(), 10, SMALL_BYTES_SIZE / 2));
+		listenerTask.join();
+		clientTask.join();
+	}
 	
 	@Test
 	public void websocketSmallSendLargeResponseTest() {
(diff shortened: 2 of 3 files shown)


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -233,17 +233,14 @@ boolean isOpen() {
...
 							else if (data instanceof ByteBuffer) {
-								bytes = (ByteBuffer) data;
-							}
-							else if (data instanceof String) {
-								bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
+								bytes = deepCopyByteBuffer((ByteBuffer) data);
 							}
 							else {
 								throw new IllegalArgumentException(
-										"The data to be sent should be String, ByteBuffer or byte[], but received " + data.getClass().getSimpleName());
```
dlstucki: >String [](start = 41, length = 6)

nice catch!

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -344,6 +341,15 @@ public void onError(Session session, Throwable cause) {
 		RelayLogger.throwingException(cause, this);
 	}
 	
+	private static ByteBuffer deepCopyByteBuffer(ByteBuffer original) {
```
dlstucki: >deepCopyByteBuffer [](start = 27, length = 18)

No action needed for now, but if you ever find the need to do this same thing from another class let's put it into a helper class at that time.

ByteBufferUtil.deepCopy or similar (Util, Utility, Helper, *Ex)

dlstucki: >deepCopyByteBuffer [](start = 27, length = 18)

Since I'm still trying to wrap my brain around ByteBuffer without reading the docs for every call I wrote this and it all passed.
```
 // Test
 public void cloneByteBufferTest() {
  byte[] array = new byte[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
  ByteBuffer source = ByteBuffer.wrap(array);
  ByteBuffer clone = deepCopyByteBuffer(source);
  assertByteBufferEquals(source, clone);

  source = ByteBuffer.wrap(array, 0, array.length / 2);
  clone = deepCopyByteBuffer(source);
  assertByteBufferEquals(source, clone);

  source = ByteBuffer.wrap(array, 1, array.length - 2);
  clone = deepCopyByteBuffer(source);
  assertByteBufferEquals(source, clone);
 }

  static void assertByteBufferEquals(ByteBuffer expected, ByteBuffer actual) {
  assertEquals(expected.capacity(), actual.capacity());  
  assertEquals(expected.arrayOffset(), actual.arrayOffset());
  assertEquals(expected.hasArray(), actual.hasArray());
  assertEquals(expected.remaining(), actual.remaining());
  assertEquals(expected.limit(), actual.limit());
  assertEquals(expected.array().length, actual.array().length);
  assertNotEquals(expected.array(), actual.array());

  while (expected.hasRemaining() && actual.hasRemaining()) {
   byte expectedValue = expected.get();
   byte actualValue = actual.get();
   assertEquals(expectedValue, actualValue);
  }

  assertEquals(expected.hasRemaining(), actual.hasRemaining());
 }
```

dlstucki: Here's a potential way to fix these issues:
```
 private static ByteBuffer deepCopyByteBuffer2(ByteBuffer original) {
  int originalPosition = original.position();
  ByteBuffer clone = ByteBuffer.wrap(new byte[original.capacity()], original.arrayOffset(), original.limit());
  original.rewind();
  clone.put(original);

  original.position(originalPosition);
  clone.position(originalPosition);

  return clone;
  }
```

---
In reply to: [296344691](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296344691) [](ancestors = 296344691)

dlstucki: >deepCopyByteBuffer [](start = 27, length = 18)

What does clone result in being the same?
- limit(), capacity() (already working)
- remaining() being the same as the original ByteBuffer _BEFORE_ it was passed to deepCopy? (it gets rewound right now).
- position() being the same as the original ByteBuffer _BEFORE_ it was passed to deepCopy? (it gets rewound right now).
- arrayOffset? (this comes from using ByteBuffer.slice().)

dlstucki: >deepCopyByteBuffer [](start = 27, length = 18)

Is deepCopyByteBuffer allowed to mutate the original ByteBuffer by changing the the current of position? What about mark?

dlstucki: >deepCopyByteBuffer [](start = 27, length = 18)

If the original ByteBuffer has consumed some bytes (e.g. position > 0) should the clone result in a ByteBuffer that when asked for the next bytes continues at the position of the original ByteBuffer or represent the original ByteBuffer after rewind was called?

dlstucki: If I call ClientWebSocket/HybridConnectionStream send/write method which takes a ByteBuffer with position != 0 will the library rewind my ByteBuffer and send bytes which I basically just told it not to?

---
In reply to: [296396267](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296396267) [](ancestors = 296396267)

dlstucki: It's possible I went a bit above and beyond what may be functionally needed. Taking a step back I created this PR using ByteBuffers with our public API.

[https://github.com/Azure/azure-relay-java/pull/49](https://github.com/Azure/azure-relay-java/pull/49)

It would be awesome to see if that new test I added works with iteration 1 of deepCopyByteBuffer.

---
In reply to: [296371319](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296371319) [](ancestors = 296371319,296344691)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -344,6 +341,15 @@ public void onError(Session session, Throwable cause) {
...
+		ByteBuffer clone = ByteBuffer.allocate(original.capacity());
+		clone.put(original.array());
```
dlstucki: >riginal.array() [](start = 13, length = 15)

Just curious, did you consider not copying the insignificant bytes?

bainian12345: I was just about to leave a comment, in that case I would probably even omit this deepcopy method and just only extract the array segment that matters. Had this here for now mostly for the method name.

---
In reply to: [296936087](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296936087) [](ancestors = 296936087)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -349,4 +375,11 @@ private static void sendResponseMessages(RelayedHttpListenerContext context, boo
 		}
 		return bytes;
 	}
+	
+	/** Validates if the ByteBuffer to be sent was modified after being sent **/
+	private static void checkMsgToSendIsModified(ByteBuffer msgToSend, ByteBuffer afterSend) {
```
dlstucki: >checkMsgToSendIsModified [](start = 21, length = 24)

Shouldn't the position move to limit as a result of being sent?

bainian12345: It should, but since we are deepcopying the buffer, the original is untouched during sending. Unless we manually change the position/remaining to simulate the effects

---
In reply to: [296937569](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296937569) [](ancestors = 296937569)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -228,29 +240,43 @@ public void httpWriteResponseAfterFlushTimerTest() throws IOException {
 	}
 	
 	private static CompletableFuture<Void> sendAndReceiveWithWebsocketClient(byte[] msgExpected, byte[] msgToSend) {
+		return sendAndReceiveWithWebsocketClient(ByteBuffer.wrap(msgExpected.clone()), ByteBuffer.wrap(msgToSend.clone()));
```
dlstucki: >clone() [](start = 71, length = 7)

these clones should no longer be needed.

src/test/java/com/microsoft/azure/relay/TestUtil.java
```diff
@@ -1,6 +1,8 @@
 package com.microsoft.azure.relay;
 
 import java.net.URI;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
```
dlstucki: >ByteBuffer [](start = 16, length = 10)

sorry about these.

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
dlstucki: I realize I have been sending mixed signals here. A general method named deepCopy probably shouldn't consume the original, but a send method which internally does a deep copy to work-around a bad behavior in the WebSocket class probably should mark the ByteBuffer that is passed in as being consumed.

---
In reply to: [296938073](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296938073) [](ancestors = 296938073,296937569)

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -228,29 +240,43 @@ public void httpWriteResponseAfterFlushTimerTest() throws IOException {
...
-			}).thenRun(() -> assertTrue("Did not receive message from websocket sender.", receivedReply.get()));
+			}).thenRun(() -> {
+				checkMsgToSendIsModified(origMsgToSend, msgToSend);
+				assertTrue("Did not receive message from websocket sender.", receivedReply.get());
+			});
 		});
 	}
 	
 	private static CompletableFuture<Void> sendAndReceiveWithWebsocketListener(byte[] msgExpected, byte[] msgToSend) {
+		return sendAndReceiveWithWebsocketListener(ByteBuffer.wrap(msgExpected.clone()), ByteBuffer.wrap(msgToSend.clone()));
```
dlstucki: >clone() [](start = 73, length = 7)

This clone can go away, I think.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -231,24 +231,23 @@ boolean isOpen() {
...
+								bytes = ((byte[]) data).clone();
 							} 
 							else if (data instanceof ByteBuffer) {
-								bytes = (ByteBuffer) data;
-							}
-							else if (data instanceof String) {
-								bytes = ByteBuffer.wrap(data.toString().getBytes(StringUtil.UTF8));
+								ByteBuffer buffer = (ByteBuffer) data;
+								bytes = new byte[buffer.remaining()];
+								buffer.get(bytes);
```
dlstucki: Sweet. Maybe ad a comment here about whatever buffer we pass being messed up by the remote.sendBinary call? Otherwise some well-meaning individual may "optimize" this extra copy out in the future.

dlstucki: Oh, it's already there lower in this method.

---
In reply to: [296943483](https://github.com/Azure/azure-relay-java/pull/48#discussion_r296943483) [](ancestors = 296943483)

PR #49, opened 2019-06-21
Add a test which uses ByteBuffers with position != 0 and remaining != capacity.

dlstucki:
Will be helpful to ensure deepCopy is correct.

### DIFF
--- a/src/test/java/com/microsoft/azure/relay/TestUtil.java
+++ b/src/test/java/com/microsoft/azure/relay/TestUtil.java
@@ -28,6 +28,7 @@ public class TestUtil {
     }
     
     static byte[] byteBufferToArray(ByteBuffer buffer) {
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/azure/relay/TestUtil.java
```diff
@@ -23,5 +25,10 @@
             offset += array.length;
         }
         return result;
-	}
+    }
+    
+    static byte[] byteBufferToArray(ByteBuffer buffer) {
```
bainian12345: >byteBufferToArray(ByteBuffer buffer) [](start = 18, length = 36)

I think `buffer1.equals(buffer2)` can replace this, since it checks for equality of the array between `position()` and `limit()` (the "remaining" segment of the buffer)

PR #54, opened 2019-10-24
TokenRenewer Fix

bainian12345:
- Prevent timer from creating multiple threads
- Prevent potential race between scheduled token renew and getTokenAsync()

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/TokenRenewer.java
+++ b/src/main/java/com/microsoft/azure/relay/TokenRenewer.java
@@ -5,13 +5,17 @@ import java.time.Instant;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
 class TokenRenewer {
 	private final HybridConnectionListener listener;
 	private final String appliesTo;
 	private final Duration tokenValidFor;
+	private final Lock scheduleLock = new ReentrantLock();
 	private Timer renewTimer;
+	private TimerTask renewTimerTask;
 	private Consumer<SecurityToken> onTokenRenewed;
 
 	protected TokenRenewer(HybridConnectionListener listener, String appliesTo, Duration tokenValidFor) {
@@ -33,7 +37,7 @@ class TokenRenewer {
 	}
 
 	private CompletableFuture<SecurityToken> getTokenAsync(boolean raiseTokenRenewedEvent) {
-			RelayLogger.logEvent("getTokenStart", this.listener);
+	    RelayLogger.logEvent("getTokenStart", this.listener);
 
 		return this.listener.getTokenProvider()
 			.getTokenAsync(this.appliesTo, this.tokenValidFor)
@@ -49,7 +53,10 @@ class TokenRenewer {
 	}
 
 	protected void close() {
-		this.renewTimer.cancel();
+	    if (this.renewTimer != null) {
+	        this.renewTimer.cancel();
+	        this.renewTimer.purge();
+	    }
 	}
 
 	void onRenewTimer() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/azure/relay/TokenRenewer.java
```diff
@@ -61,23 +68,38 @@ void onRenewTimer() {
...
 	private void scheduleRenewTimer(SecurityToken token) {
-		this.renewTimer = new Timer();
-		Duration interval = Duration.between(Instant.now(), token.getExpiresAtUtc());
-		if (interval.isNegative()) {
-			RelayLogger.logEvent("tokenRenewNegativeDuration", this.listener);
-			return;
-		}
+	    if (this.renewTimer == null) {
+	        this.renewTimer = new Timer();    
+	    }
```
dlstucki: Is it preferable to have this outside the scheduleLock protected code? Would it be better to put this inside the lock?

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
```diff
@@ -5,13 +5,17 @@
...
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
 class TokenRenewer {
 	private final HybridConnectionListener listener;
 	private final String appliesTo;
 	private final Duration tokenValidFor;
+	private final Lock scheduleLock = new ReentrantLock();
```
dlstucki: >ReentrantLock [](start = 39, length = 13)

Is there a reason to use ReentrantLock instead of synchronized(this) or whatever?
``` JAVA
public class DemoClass
{
    public synchronized void demoMethod(){}
}
 
or
 
public class DemoClass
{
    public void demoMethod(){
        synchronized (this)
        {
            //other thread safe code
        }
    }
}
```
[https://howtodoinjava.com/java/multi-threading/object-vs-class-level-locking/](https://howtodoinjava.com/java/multi-threading/object-vs-class-level-locking/)

dlstucki: From [https://howtodoinjava.com/java/multi-threading/how-to-use-locks-in-java-java-util-concurrent-locks-lock-tutorial-and-example/](https://howtodoinjava.com/java/multi-threading/how-to-use-locks-in-java-java-util-concurrent-locks-lock-tutorial-and-example/):

Difference between Lock Interface and synchronized keyword
The main differences between a Lock and a synchronized block are:

1) Having a timeout trying to get access to a synchronized block is not possible. Using Lock.tryLock(long timeout, TimeUnit timeUnit), it is possible.
2) The synchronized block must be fully contained within a single method. A Lock can have it’s calls to lock() and unlock() in separate methods.

---
In reply to: [338833472](https://github.com/Azure/azure-relay-java/pull/54#discussion_r338833472) [](ancestors = 338833472)

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
```diff
@@ -61,23 +68,38 @@ void onRenewTimer() {
...
-		}, interval.toMillis());
+	            if (this.renewTimerTask != null) {
+	                this.renewTimerTask.cancel();
+	            }
+                this.renewTimerTask = new TimerTask() {
+                    @Override
+                    public void run() {
+                        onRenewTimer();
+                    }
+                };
```
dlstucki: Why must you cancel the old and create again? Would creating it only the first time work?
``` JAVA
             if (this.renewTimerTask == null) {
                this.renewTimerTask = new TimerTask() {
                    @Override
                    public void run() {
                        onRenewTimer();
                    }
                };
             }
```

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
bainian12345: Don't think it would make a difference in this case in or out of the locked block

---
In reply to: [338832149](https://github.com/Azure/azure-relay-java/pull/54#discussion_r338832149) [](ancestors = 338832149)

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
bainian12345: TimerTask cannot be reused, we have to create a new object each time. Cancel is in case if we got a new token and renewed before the timer, we probably want to avoid the task from executing since we already have another one scheduled

---
In reply to: [338835245](https://github.com/Azure/azure-relay-java/pull/54#discussion_r338835245) [](ancestors = 338835245)

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
bainian12345: I mostly wanted ```.tryLock()```, which just skips the block instead of ```synchronized()```, which sits there and block while wait for the lock

---
In reply to: [338833472](https://github.com/Azure/azure-relay-java/pull/54#discussion_r338833472) [](ancestors = 338833472)

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
dlstucki: Please put it inside the block.

---
In reply to: [338835499](https://github.com/Azure/azure-relay-java/pull/54#discussion_r338835499) [](ancestors = 338835499,338832149)

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
```diff
@@ -61,23 +68,38 @@ void onRenewTimer() {
...
 
-		RelayLogger.logEvent("tokenRenewScheduled", this.listener, interval.toString());
-		this.renewTimer.schedule(new TimerTask() {
-			@Override
-			public void run() {
-				onRenewTimer();
-			}
-		}, interval.toMillis());
+	            if (this.renewTimerTask != null) {
+	                this.renewTimerTask.cancel();
```
dlstucki: >this.renewTimerTask.cancel(); [](start = 17, length = 29)

Would you consider putting a comment for simpletons like me to understand this? Something like `// Cancel any previously scheduled task`

src/main/java/com/microsoft/azure/relay/TokenRenewer.java
dlstucki: Ah, I see.

---
In reply to: [338836278](https://github.com/Azure/azure-relay-java/pull/54#discussion_r338836278) [](ancestors = 338836278,338833472)

PR #58, opened 2019-12-03
Fix listener reconnection + Fault Injection

bainian12345:


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -3,7 +3,9 @@
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.time.Duration;
+import java.util.HashSet;
```
bainian12345: These imports aren't needed, will remove

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -289,28 +290,46 @@ public void parseRequestPathAndQueryTest() throws IOException, URISyntaxExceptio
...
-		assertEquals("The reconnecting handler was not called exactly once", 1, handlerExecuted.get());
+		assertEquals("The reconnecting handler was not called exactly once.", 1, handlerExecuted.get());
 		assertTrue("Exception should be ConnectionLostException.", exception.get() instanceof ConnectionLostException);	
 		reconnectTask.join();
 	}
 	
+	// The listener should continuously try to reconnect if network is lost
+	@Test
+	public void listenerNetworkLostTest() throws Exception {
+	    int delayIndex = 2;
```
bainian12345: >2 [](start = 22, length = 1)

Should work for all values 2-5, I just wanted the test to take the minimal time

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -339,8 +352,14 @@ public void onClose(Session session, CloseReason reason) {
 	}
 
 	@OnError
-	public void onError(Session session, Throwable cause) {
-		RelayLogger.throwingException(cause, this);
+	public void onError(Throwable cause) {
```
dlstucki: >public void onError(Throwable cause) [](start = 1, length = 36)

I see the onError signature changed. Are both of these valid (or at least the latest one)?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -494,15 +491,43 @@ private URI generateAcceptRequestUri(URI rendezvousUri) throws URISyntaxExceptio
...
+	}
+	
+	/**
+	 * Inject a specific exception to the listener or its underlying websocket for testing purposes
+	 * @param throwable The following exceptions can be interpreted:
+	 * </br><b>UpgradeException</b>    - When this exception is injected, all listener websocket connection attempts will fail 
+	 *                                   until the fault has been cleared.
+	 * </br><b>ConnectionException</b> - Injecting this exception will kill the listener's websocket (simulate a network disconnect),
+	 *                                   but should cause the listener to try to reconnect.
+	 */
```
dlstucki: Cool stuff!

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -696,6 +714,10 @@ else if (command.getRenewToken() != null) {
 
 				ClientWebSocket webSocket = new ClientWebSocket(this.listener.trackingContext, EXECUTOR);
 				return delayTask.thenCompose(($void) -> {
+	                if (this.listener.injectedFault != null && this.listener.injectedFault instanceof UpgradeException) {
```
dlstucki: >this.listener.injectedFault instanceof UpgradeExceptio [](start = 60, length = 54)

injectedFault seems to only ever get set if the exception is-an UgradeException. With the current code I can't see any way for this instanceof check to ever return false.

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
bainian12345: It should not make a difference, both are valid. I haven't observed any difference either.

---
In reply to: [354519842](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354519842) [](ancestors = 354519842)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -36,6 +37,7 @@
 	private URI address;
 	private TrackingContext trackingContext;
 	private TokenProvider tokenProvider;
+	private Throwable injectedFault;
```
dlstucki: >injectedFault [](start = 19, length = 13)

Nit: If this is only ever used/throw during connect/re-connect could the name be updated to indicate that?

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
bainian12345: If there are more faults needed in the future we just add here. I wonder if there are better ways to doing this though.

---
In reply to: [354520748](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354520748) [](ancestors = 354520748)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: >Throwable [](start = 9, length = 9)

Given the current code you could easily change this to just be UpgradeException. (would require one cast after the instanceof UgradeException in the `injectFault` method.

bainian12345: For now that's the only possible one, I felt like it could be reused if there are more in the future

---
In reply to: [354522333](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354522333) [](ancestors = 354522333)

src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
dlstucki: As long as it gets called it's good.

---
In reply to: [354522105](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354522105) [](ancestors = 354522105,354519842)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: Sounds like you want to keep it generic for future use. ok.

---
In reply to: [354522976](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354522976) [](ancestors = 354522976)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: Sounds like you want to keep it generic for future use. ok. Close/resolve/whatever

---
In reply to: [354521815](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354521815) [](ancestors = 354521815)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -289,28 +290,46 @@ public void parseRequestPathAndQueryTest() throws IOException, URISyntaxExceptio
...
 		});
 		
 		listener.openAsync(Duration.ofSeconds(15)).join();
-		ControlConnection controlConnection = listener.getControlConnection();
-		controlConnection.getConnectAsyncTask().thenCompose(controlWebSocket -> {
-			return controlWebSocket.closeAsync(); // This bypasses listener.close(), simulating an unexpected close
-		}).join();
-		
+		listener.injectFault(new ConnectionLostException());
+		assertFalse("listener should be disconnected temporarily for now", listener.isOnline());
```
dlstucki: >assertFalse [](start = 2, length = 11)

Is there ever a race between injectFault causing the websocket to get disposed and this line? I think if the `controlConnection.connectAsyncTask` is already completed then when it calls `.thenApply` to dispose the websocket that might happen immediately? (or will that run it async?)

bainian12345: Theoretically there is most definitely a race. However, disposing the websocket is a VERY LONG operation (takes 10 sec or more), I didn't want to be left hanging for that long, and isOnline() will return false when the call is made, not necessarily after the call finishes.

---
In reply to: [354528043](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354528043) [](ancestors = 354528043)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
dlstucki: Not needed for this check-in but I'm curious if there are ways (events or similar) to make it catch the online handler as soon as it happens instead of sleeping long enough that it should happen no matter what...

---
In reply to: [354519489](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354519489) [](ancestors = 354519489)

src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
bainian12345: Bottom line is, I feel that as long as the websocket is closing and will get disposed, I don't think the user cares about how long it will take.

---
In reply to: [354529498](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354529498) [](ancestors = 354529498,354528043)

src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
dlstucki: In the future maybe have an enum specifying when to use the exception plus an exception. That would allow throwing the same exception type from different locations, for example.

---
In reply to: [354522602](https://github.com/Azure/azure-relay-java/pull/58#discussion_r354522602) [](ancestors = 354522602,354520748)

Issue #60, opened 2020-01-26
ListenerCommand improper JSON serialisation


consultantleon:
We found that the code to serialize the ListenerCommand toJsonString() method does not do a proper JSON serialization.
If any of the values serialized contains \ or " characters (or other special characters that JSON values should escape), the result is an invalid JSON serialization.
I'd suggest not to use JSONObject in the code (seems not to be a JSON de/serializer of of choice), but instead use proven JSON mapper solution from jackson/fasterxml.
At least never attempt to d.i.y. strings to JSON as it easily goes wrong as our case proved.

In below code we ran into troubles if HTTP headers contain a " character.

```
String toJsonString() {
			StringBuilder builder = new StringBuilder("{\"" + HybridConnectionConstants.Actions.RESPONSE + "\":{");
			List<String> fields = new ArrayList<String>();

			if (this.requestId != null) {
				fields.add("\"requestId\":\"" + this.requestId + "\"");
			}
			
			fields.add("\"body\":" + ((this.body) ? "true" : "false"));
			fields.add("\"statusCode\":" + this.statusCode);

			if (this.statusDescription != null) {
				fields.add("\"statusDescription\":\"" + this.statusDescription + "\"");
			}

			if (this.responseHeaders != null && !this.responseHeaders.isEmpty()) {
				StringBuilder headerBuilder = new StringBuilder("\"responseHeaders\":{");
				this.responseHeaders.forEach((key, val) -> {
					headerBuilder.append("\"" + key + "\":\"" + val + "\",");
				});
				headerBuilder.replace(headerBuilder.length() - 1, headerBuilder.length(), "}");
				fields.add(headerBuilder.toString());
			}
			
			builder.append(String.join(",", fields)).append("}}");
			return builder.toString();
		}
	}
```

We've done a quick internal patch to work around this bug using the JSONObject to do the serialization:
```
 if (this.responseHeaders != null && !this.responseHeaders.isEmpty()) {
                StringBuilder headerBuilder = new StringBuilder("\"responseHeaders\":");
                headerBuilder.append(new JSONObject(responseHeaders).toString());
                fields.add(headerBuilder.toString());
            }
```
(still we'd prefer to see this code use jackson / fasterxml)

jfggdl:
@consultantleon Thank you for reporting this issue. We will work on this bug. The fix will probably not use external libraries to avoid additional dependencies.

consultantleon:
[...]

PR #62, opened 2020-01-29
Fix encoding of special characters in ListenerCommand json

bainian12345:
- Fix json encoding of characters such as "
- Updated test for  custom headers

### DIFF
--- a/src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
+++ b/src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
@@ -3,6 +3,7 @@ package com.microsoft.azure.relay;
 import static org.junit.Assert.*;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -76,42 +75,66 @@ public void openAndCloseTest() {
...
+
+        Map<String, List<String>> headers = httpConnection.getHeaderFields();
+        assertNotNull(headers);
+        assertEquals(httpConnection.getResponseMessage(), 200, responseCode);
+        httpRequestReceived.get(10, TimeUnit.SECONDS);
+
+        // Test sending custom header from client and receiving from listener for HybridConnection websocket mode
+        Map<String, List<String>> customHeaders = new HashMap<String, List<String>>();
+        customHeaders.put(allowedHeaderNameChars, Arrays.asList(new String[]{allowedHeaderValChars}));
+		CompletableFuture.allOf(client.createConnectionAsync(customHeaders).thenCompose(connection -> {
```
dlstucki: >		 [](start = 0, length = 2)

Nit: The shows as not indented like the other surrounding lines in some editors because some lines use tabs and some lines use spaces.

consultantleon:
this is basically the fix for the issue I raised #60 , good to update the status on #60 as fixed in release 0.0.3! and link to this issue.

PR #68, opened 2020-03-23
Improve synchronization of fields

bainian12345:
Improve synchronization of fields in HybridConnectionListener.java

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
+++ b/src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
@@ -11,6 +11,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/azure/relay/HybridConnectionListener.java
```diff
@@ -531,7 +538,7 @@ void injectFault(Throwable throwable) {
 	 * Connects, maintains, and transparently reconnects this listener's control
 	 * connection with the cloud service.
 	 */
-	static final class ControlConnection implements AutoCloseable {
+	final class ControlConnection implements AutoCloseable {
 		private final HybridConnectionListener listener;
```
dlstucki: >private final HybridConnectionListener listener; [](start = 2, length = 48)

It's a minor point and might introduce unnecessary churn but once this class is no longer static you no longer need to explicitly pass in the "listener" reference, right?

bainian12345: In that case the "listener" reference would have to be referred to as "HybridConnectionListener.this", I personally feel like it's more readable as it is currently.

---
In reply to: [397338730](https://github.com/Azure/azure-relay-java/pull/68#discussion_r397338730) [](ancestors = 397338730)

PR #69, opened 2020-03-24
make status description for HTTP response be JSON safe

bainian12345:
Make status description for HTTP response be JSON safe with same approach used on the custom response headers.


src/test/java/com/microsoft/azure/relay/HybridConnectionListenerTest.java
```diff
@@ -87,6 +87,7 @@ public void customHeadersTest() throws Exception {
 	        try {
 	            Map<String, String> headers = context.getRequest().getHeaders();
 	            assertEquals(allowedHeaderValChars, headers.get(allowedHeaderNameChars));
+	            context.getResponse().setStatusDescription(allowedHeaderValChars);
```
dlstucki: >setStatusDescription [](start = 35, length = 20)

Would it be possible to update this test to check this StatusDescription makes it properly to the HTTP client?

PR #81, opened 2022-02-14
Enables text send and read operations for WebSocket connections. 

### LINKED ISSUE #80: Can't write or read text data using the rendezvous connection
There isn't seem a way to write text data to the rendezvous connection.

```listener.acceptConnectionAsync().join()``` returns  ```HybridConnectionChannel```, which I can cast to ```WebSocketChannel``` as that is the underlying type. However, the method I need,  ```CompletableFuture<Void> writeAsync(Object data, Duration timeout, boolean isEnd, WriteMode mode)```, is not public. 

Similarly, the ``readTextAsync``` method is not public, hence there is no way to read text messages.

### PR DESCRIPTION
**Pull request by giventocode**
Closes #80

### DIFF
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -156,7 +156,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	 * 
 	 * @return Returns a CompletableFuture which completes when websocket receives text messages
 	 */
-	public CompletableFuture<String> readTextAsync() {
+	CompletableFuture<String> readTextAsync() {
 		return this.textQueue.dequeueAsync().thenApply(text -> {
 			if (text != null) {
 				RelayLogger.logEvent("receivedText", this, String.valueOf(text.length()));
@@ -219,7 +219,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	 * @return A CompletableFuture which completes when websocket finishes sending the bytes.
 	 * @throws TimeoutException Throws when the sending task does not complete within the given timeout.
 	 */
-	public CompletableFuture<Void> writeAsync(Object data, Duration timeout, boolean isEnd, WriteMode mode) {
+	CompletableFuture<Void> writeAsync(Object data, Duration timeout, boolean isEnd, WriteMode mode) {
 		if (this.isOpen()) {
 			if (data == null) {
 				// TODO: Log warns sending nothing because message is null
@@ -330,6 +330,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 				textQueue.enqueueAndDispatch(text);
 			}
 		});
+
 
 		session.addMessageHandler(new MessageHandler.Partial<byte[]>() {
 			@Override
--- a/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
+++ b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
@@ -156,7 +156,7 @@ class ClientWebSocket extends Endpoint implements RelayTraceSource {
 	 * 
 	 * @return Returns a CompletableFuture which completes when websocket receives text messages
 	 */
-	CompletableFuture<String> readTextAsync() {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/main/java/com/microsoft/azure/relay/ClientWebSocket.java [hunk 1]
```diff
@@ -156,7 +156,7 @@ boolean isOpen() {
 	 * 
 	 * @return Returns a CompletableFuture which completes when websocket receives text messages
 	 */
-	CompletableFuture<String> readTextAsync() {
+	public CompletableFuture<String> readTextAsync() {
```
bainian12345, line 159: >public C [](http://example.com/codeflow?start=1&length=8)

Why do you need to make the two methods in this class public? `WebSocketChannel` is more public facing and `ClientWebSocket` is more of an internal implementation class, so I would prefer if you leave these methods internal.

bainian12345: You can expose public methods in `WebSocketChannel` class instead.

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -73,6 +74,16 @@ public void websocketSmallSendSmallResponseTest() {
...
+	public void websocketSendTextResponseTextTest() {
+		CompletableFuture<Void> listenerTask = sendAndReceiveTextWithWebsocketListener(WS_MSG, WS_MSG);
+		CompletableFuture<Void> clientTask = sendAndReceiveTextWithWebsocketClient(WS_MSG, WS_MSG);
+				
+		listenerTask.join();
+		clientTask.join();
+	}
+
+
 	@Test
```
bainian12345, line 87: nit: remove the extra empty line

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
```diff
@@ -73,6 +74,16 @@ public void websocketSmallSendSmallResponseTest() {
 		clientTask.join();
 	}
 
+	@Test
+	public void websocketSendTextResponseTextTest() {
+		CompletableFuture<Void> listenerTask = sendAndReceiveTextWithWebsocketListener(WS_MSG, WS_MSG);
```
bainian12345, line 79: I would like to see larger data (>64kb) be tested as well. Maybe you can create small and large strings out of the `SMALL_BYTES` and `LARGE_BYTES` from above.

src/main/java/com/microsoft/azure/relay/WebSocketChannel.java
```diff
@@ -113,7 +113,17 @@ public void close() throws IOException {
...
-	
+	/**
+	 * Sends the text data to the remote endpoint within a timeout in one of the WriteModes.
+	 *
+	 * @param textData Text message to be sent.
+	 * @param timeout The timeout to connect to send the data within. May be null to indicate no timeout limit.
+	 * @return A CompletableFuture which completes when websocket finishes sending the data.
+	 * @throws TimeoutException Throws when the sending task does not complete within the given timeout.
+	 */
+	public CompletableFuture<Void> writeTextAsync(String textData, Duration timeout) {
```
bainian12345, line 124: Just wondering, is there any functional difference where you would need to use WriteMode.TEXT instead?

bainian12345 (changes requested):
:clock1:

src/test/java/com/microsoft/azure/relay/SendReceiveTest.java
giventocode: >64kb is not possible in a single message. The `DefaultMaxTextMessageBufferSize` is by default `RelayConstants.DEFAULT_CONNECTION_BUFFER_SIZE`, which is 64Kb. I added a test with a string of 64KB.

src/main/java/com/microsoft/azure/relay/WebSocketChannel.java
giventocode: Perhaps if you want to control the encoding of the String. However, in our case there is no value in exposing the WriteMode or multi-part messages.

bainian12345 (approved):
:shipit:

PR #88, opened 2022-11-01
.env, logging, and updated dependencies

tvenhaus:
This Pull Request implements the following changes:

1. Introduces a .env file to enable test execution without updating a system environment variable
2. Updates dependencies to the latest versions
3. Makes log4j a dependency only for the "test" scope

The changes to the `.classpath`, `.project`, and `pom.properties` files are made by VSCode when the project is opened and compiled.  If these changes have a negative impact to other IDEs, we will need to reconcile.


pom.xml
```diff
@@ -31,20 +31,28 @@
...
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<azure.identity.version>1.4.0</azure.identity.version>
-		<jetty.version>9.4.44.v20210927</jetty.version>
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<maven.compiler.source>1.8</maven.compiler.source>
-		<slf4j-version>1.7.0</slf4j-version>
-		<log4j-version>1.2.17</log4j-version>
+		<azure.identity.version>1.6.1</azure.identity.version>
+		<slf4j.version>2.0.3</slf4j.version>
+		<jetty.version>9.4.49.v20220914</jetty.version>
```
pjlittle, line 38: to double check, these are pegged to the latest version?

.project
```diff
@@ -20,4 +20,15 @@
...
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
+		<filter>
+			<id>1667084030105</id>
+			<name></name>
+			<type>30</type>
+			<matcher>
+				<id>org.eclipse.core.resources.regexFilterMatcher</id>
+				<arguments>node_modules|\.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
```
pjlittle, line 30: is this specific to Eclipse or used in the build process? Does VSCode or similar leverage this file?

pjlittle: nvm - saw what you mentioned in the description. So Code uses this as well;

pom.xml
tvenhaus: Yes

pjlittle (approved):

PR #89, opened 2022-11-09
Replace JSR-356 WebSocket implementation with Jetty

tvenhaus:
The Jetty WebSockets package (`org.eclipse.jetty.websocket`) has several advantages over the JSR-356 package (`javax.websockets`) including (but not limited to)...
- Better support for SSL/TLS
- Support for HTTP proxies


jbjordan:
All tests passed

jbjordan (approved):

PR #91, opened 2022-12-09
Implement keep-alive via WebSocket Ping

tvenhaus:
WebSocket connections are often meant to be open for long periods of time.  The purpose of Ping messages is to keep networking appliances from closing the connection due to a period of inactivity.  

This Pull Request will result in all WebSocket connections sending a Ping message every 30 seconds.


src/main/java/com/microsoft/azure/relay/ClientWebSocket.java
```diff
@@ -475,4 +478,20 @@ private CompletableFuture<Void> readFragmentsAsync() {
 					});
 		}
 	}
+	
+	private class PingRunnable implements Runnable {
```
tasdevani21, line 482: Would there be any consideration to split this private class into another file vs keeping it here? and are there any advantages/disadvantages of doing it one way?

jbjordan: I think it makes sense to keep it nested since this functionality is not used anywhere else (and there are no plans for this) and it uses some of ClientWebSocket's private variables.

jbjordan (commented):
All test cases passed

jbjordan (approved):

